PR #12, opened 2016-07-04
DAT-4523 | Build SVG sprite

rogatty:
https://wikia-inc.atlassian.net/browse/DAT-4523

- add gulp task for building `dist/symbols.svg`
- deduplicate ids in SVG files
- move `index.scss` to `dist/styles.scss`


hakubo:
LGTM

gulpfile.js
```diff
@@ -0,0 +1,39 @@
+var gulp = require('gulp'),
```
hakubo: should we use https://github.com/Wikia/javascript-style-guide here?

rogatty: Good question. We can't use it fully because it requires ES6 and I don't think it's worth to implement Babel just for the gulpfile (wow, that would be beautifully over-engineered). And I think we shouldn't depend on node.js version being > 4 before we have it updated on all devboxes.

PR #14, opened 2016-07-11
DAT-4600 Add fandom default link color

kamilkoterba:
https://wikia-inc.atlassian.net/browse/DAT-4600
@Wikia/x-wing


kamilkoterba:
Looks simple, right? :)

vforge:
LGTM

PR #15, opened 2016-07-11
DAT-4613 | add global-footer styles

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/DAT-4613


dist/styles.scss
```diff
@@ -1 +1,2 @@
 @import '../base/base';
+@import '../components/components';
```
rogatty: `../components/index`?

tomasznapieralski: can we also have /base/index? it is confusing

vforge:
LGTM

PR #20, opened 2016-07-25
Add assets/* and gulpfile.js to the list of files ignored by bower

rogatty:
106 less unnecessary files to install.


tomasznapieralski:
LGTM

PR #22, opened 2016-07-26
XW-1756 | add icon styles + add icon prefix

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/XW-1756

@Wikia/x-wing


rogatty:
LGTM

PR #23, opened 2016-07-26
XW-1757 | GlobalFooter allow longer translations

kvas-damian:
https://wikia-inc.atlassian.net/browse/XW-1757


components/_global-footer.scss
```diff
@@ -463,31 +448,37 @@
...
 				display: block;
+				order: 5;
 				width: $desktop-third-column-width;
 			}
 
 			&.wds-is-community-apps {
-				height: $desktop-row-top-height;
+				margin-right: 0;
+				min-height: $desktop-row-top-height;
+				order: 3;
```
hakubo: why do we need this ordering here?

hakubo: actually why do we need all order properties?

components/_global-footer.scss
```diff
@@ -497,7 +488,8 @@
 			}
 
 			&.wds-is-advertise {
-				height: $desktop-row-bottom-height;
+				margin-right: 0;
```
hakubo: code smell...
is that to override some styles? or set default?

components/_global-footer.scss
kvas-damian: On mobile we display sections one under another. On desktop I decided to take another approach and display them one next to another and break row after 3rd.

The reason why I decided to change flex-direction:
* Previously on desktop we tried to emulate the same flow by changing flex-direction to columns, then we had to limit height of the container to force flex to wrap column. Finally we end up with static height containers.
* css column doesn't allow to manipulate column widths
* floats won't work, as content is in different order

Other option is to add additional HTML markup only for styling columns, which I personally don't like.

components/_global-footer.scss
kvas-damian: Override defaults. Maybe we could convert those margins into `justify-content: space-between` although I'm not sure if it should be done as a part of this ticket.

tomasznapieralski:
Tested on Chrome, FF, IE, Edge everywhere is good, more in ticket :)

PR #25, opened 2016-08-03
XW-1799 | Build individual SVGs

rogatty:
https://wikia-inc.atlassian.net/browse/XW-1799

We need to build individual SVGs so they can be styled when inlined in the HTML. This PR:
- remove `fill` attribute in order to allow changing colors of svgs. Previously we used different technique which adds unnecessary complexity.
- moves `/dist/*.svg` to `/dist/svg/*.svg` because now there is a lot of them
- renames `company.svg` to `sprite-company.svg` and `icons.svg` to `sprite-icons.svg` to better reflect what they are
- adds gulp task that builds individual SVGs
  - name them like `wds-icons-facebook.svg` so we don't need [mapping in the app](https://github.com/Wikia/app/blob/58504e17c775dc61da7be2c81c1c7a01c0e09698/extensions/wikia/DesignSystem/DesignSystemHelper.class.php#L63-L77)


kvas-damian:
I'm not sure why do we need to add class to svg elements. As far as I understand this PR we would to achieve cascading fill color. Which IMO can be accomplished by just removing fill attribute and defining 
```
svg {
  fill: currentColor;
}
```

What am I missing?

rogatty:
@kvas-damian you are right, this is not needed when we inline SVGs in place. It's required only when using sprites. I fixed that in the last commit.

kvas-damian:
Probably we don't need `class=fill` after merging https://github.com/Wikia/design-system/pull/26

rogatty:
LGTM

hakubo:
LGTM

PR #28, opened 2016-08-11
XW-1856 | divider length on desktop; full site link appearance

Warkot:
Applied designer's feedback.

https://wikia-inc.atlassian.net/browse/XW-1856

@Wikia/x-wing

### DIFF
--- a/components/_global-footer.scss
+++ b/components/_global-footer.scss
@@ -88,11 +88,28 @@
 	}
 
 	&__bottom-bar-row {
+		margin: 0 auto;
 		padding: 16px 0;
 
 		&.wds-has-border-top {
 			border-top: 1px solid $color-slate-gray;
 		}
+
+		&.wds-has-no-padding {
+			padding: 0;
+		}
+	}
+
+	// @todo To be moved to a separate component
+	&__button-link {
+		display: inline-block;
+		font-weight: bold;
+		letter-spacing: .3px;
+		line-height: normal;
+		min-height: 45px;
+		padding: 14px 18px;
+		text-decoration: none;
+		text-transform: uppercase;
 	}
 
 	&__fandom-logo {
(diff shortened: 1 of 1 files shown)


components/_global-footer.scss
```diff
@@ -88,11 +88,28 @@
...
+		&.wds-has-no-padding {
+			padding: 0;
+		}
+	}
+
+	// @todo To be moved to a separate component
+	&__button-link {
+		display: inline-block;
+		font-weight: bold;
+		letter-spacing: .3px;
```
hakubo: that should be defined in global styles, should it?
why do we define it only for button link?

components/_global-footer.scss
```diff
@@ -88,11 +88,28 @@
...
 
 	&__bottom-bar-row {
+		margin: 0 auto;
 		padding: 16px 0;
 
 		&.wds-has-border-top {
 			border-top: 1px solid $color-slate-gray;
 		}
+
+		&.wds-has-no-padding {
```
hakubo: if we  did the opposite we would not have to override padding property.

components/_global-footer.scss
Warkot: I copied it from the "Start a Wikia" button and we should define it for all buttons as it's defined here https://zpl.io/aohS8, when creating a button component in DS.

components/_global-footer.scss
Warkot: My intuition told me that the default row should be padded. The row with just a button seemed like an custom exception (yeah, I know there are only two rows 😝 ). And the option you're proposing might look like that:
```
	&__bottom-bar-row {
		margin: 0 auto;
		…
		&.wds-has-padding {
			padding: 16px 0;
		}
	}
```
…and it actually looks pretty damn good, now when I look at it. 👍

Warkot:
LGTM from @hakubo ;)

rogatty:
LGTM

PR #29, opened 2016-08-11
XW-1856 | fix line breaking for CJK languages (Chinese, Japanese, and Korean)

Warkot:
https://wikia-inc.atlassian.net/browse/XW-1859

@Wikia/x-wing


kvas-damian:
LGTM

Grunny:
LGTM.

Warkot:
I'm reverting this as this breaks CJK strings on Safari—they actually don't break at all even with `word-wrap: break-word;`.

PR #31, opened 2016-08-17
XW-1874 | adjust footer line height/list items spacing

Warkot:
https://wikia-inc.atlassian.net/browse/XW-1874

@Wikia/x-wing

### DIFF
--- a/components/_global-footer.scss
+++ b/components/_global-footer.scss
@@ -68,7 +68,7 @@
 	&__bottom-bar {
 		background: rgba($color-night-sky, .5);
 		font-size: $typescale-size-minus-2;
-		line-height: $line-height;
+		line-height: $footer-line-height;
 		min-height: 50px;
 		padding: 0 20px;
 		text-align: center;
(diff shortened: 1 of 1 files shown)


components/_global-footer.scss
```diff
@@ -25,9 +25,10 @@
 	$mobile-row-header-international-padding-bottom: 35px;
 
 	$section-line-width: 4px;
-	$desktop-row-top-height: 205px + $section-line-width;
+	$desktop-row-top-height: 206px + $section-line-width;
 
 	// others
+	$footer-line-height: 1.3em;
```
kvas-damian: While we're introducing new `line-height` value do you think we should use it for entire `global-footer`? I see old line height used in `&__bottom-bar` section

Warkot: Fixed. https://github.com/Wikia/design-system/pull/31/commits/0391904bbb982243da9d90130d948cfbfcf7ec01

kvas-damian:
LGTM

PR #32, opened 2016-08-19
XW-1618 | Add global-navigation initial styles + add button styles + add sample webcomponent

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/XW-1618

@Wikia/x-wing


tomasznapieralski:
LGTM

components/_buttons.scss
```diff
@@ -0,0 +1,59 @@
...
+	display: inline-block;
+	font-size: $typescale-size-minus-2;
+	font-weight: 600;
+	letter-spacing: .15px;
+	line-height: $button-line-height;
+	padding: 16px 18px;
+	text-align: center;
+	text-decoration: none;
+	text-transform: uppercase;
+	transition: all 300ms;
```
Warkot: Do we need to animate `all` or maybe just what is actually animated?

components/_buttons.scss
```diff
@@ -0,0 +1,59 @@
...
+	-webkit-appearance: none;
+	-webkit-border-radius:0;
+
+	&:hover, &.wds-active {
```
Warkot: I would break all separate selectors not only separated by a comma, but additionally with a line break. Same way it's done in `_global-footer.scss`. This way it's easier to distinguish several selectors from a long single selector.

components/_buttons.scss
```diff
@@ -0,0 +1,59 @@
...
+		background-color: $button-active-color;
+		border-color: $button-active-color;
+	}
+
+	&.disabled, &[disabled] {
+		cursor: default;
+		opacity: 0.5;
```
Warkot: No leading zeros please. https://github.com/Wikia/guidelines/blob/master/SCSS/CodingConventions.md#leadingzero

components/_global-navigation.scss
```diff
@@ -0,0 +1,123 @@
...
+$nav-height: 49px;
+
+.wds-global-navigation {
+	border-bottom: 1px solid $color-faint-gray;
+	border-top: $border-top-width solid $color-slate-gray;
+	display: flex;
+	justify-content: center;
+
+	&::before, &::after {
+		content: "";
```
Warkot: Single quotes are preffered.
https://github.com/Wikia/guidelines/blob/master/SCSS/CodingConventions.md#stringquotes

Warkot:
One medicore and three minor comments. Besides that LGTM.

components/_global-navigation.scss
```diff
@@ -0,0 +1,123 @@
...
+	&::before {
+		border-top: $border-top-width solid $color-dark-blue-gray;
+		margin-top: -$border-top-width;
+	}
+
+	&__content-wrapper {
+		align-items: center;
+		display: flex;
+		height: $nav-height;
+		width: 1024px;
```
Warkot: Don't we have a variable for that value?

components/_global-navigation.scss
```diff
@@ -0,0 +1,123 @@
...
+
+			&:active, &.active {
+				background-color: $color-tv;
+				border-color: darken($color-tv, 10%);
+			}
+		}
+	}
+
+	&__search {
+		background-color: lightgray;
```
Warkot: It doesn't look like a color from Design System. Or even as a variable. ;)

components/_global-navigation.scss
kvas-damian: I see one here: https://github.com/Wikia/design-system/blob/22833834db692ed4a3ec4cf4991eb48abc7dfec5/components/_global-footer.scss#L32
Although I don't think we should reuse it. These is one more `1024px` variable that defines breakpoint, but I'm not sure if they should be bound.

Warkot:
LGTM

webcomponents/global-navigation.html
```diff
@@ -0,0 +1,14 @@
...
+	skate.define('wds-global-navigation', {
+		render(elem) {
+			skate.vdom.text(`Hello, ${elem.name}`);
+		}
+	});
+</script>
+<template id="template">
+    <div>
+        <h1>Web Components</h1>
+        <img src="http://webcomponents.org/img/logo.svg">
```
rogatty: Why do we want to merge it to master?

components/_buttons.scss
```diff
@@ -0,0 +1,59 @@
...
+	line-height: $button-line-height;
+	padding: 16px 18px;
+	text-align: center;
+	text-decoration: none;
+	text-transform: uppercase;
+	transition: background-color 300ms, border-color 300ms, color 300ms;
+	-webkit-appearance: none;
+	-webkit-border-radius: 0;
+
+	&:hover, &.wds-active {
```
rogatty: Shouldn't this be `.wds-is-active`? Same for all modifiers.

webcomponents/global-navigation.html
tomasznapieralski: we can remove it

rogatty:
LGTM

PR #33, opened 2016-08-22
XW-1619 | add list styles

tomasznapieralski:


components/_lists.scss
```diff
@@ -0,0 +1,22 @@
+.wds-list {
+  display: block;
+  list-style-type: none;
+  padding: 0;
+  margin: 18px;
+
+  .wds-small-list-item {
```
hakubo: this should not be nested

hakubo: this should be divided into list item and a modifier

.wds-list-item

and
.wds-is-small/big

we should identify what is more commonly used and make it a default
so you don't have to set the modifier always

components/_lists.scss
```diff
@@ -0,0 +1,22 @@
...
+    display: flex;
+    flex-direction: column;
+    font-size: $typescale-size-minus-1;
+    height: 36px;
```
hakubo: if an item will have more content to fit into one line
this will break

instead of height we could use flex with min height

components/_lists.scss
```diff
@@ -0,0 +1,19 @@
+.wds-list {
+  display: block;
```
rogatty: We use tabs.

tomasznapieralski: Comment addressed in https://github.com/Wikia/design-system/pull/38

tomasznapieralski:
Closing as duplicated for https://github.com/Wikia/design-system/pull/38

PR #34, opened 2016-08-22
XW-1903 | add dropdown styles

tomasznapieralski:


components/_dropdowns.scss
```diff
@@ -0,0 +1,20 @@
+.wds-dropdown-toggle {
+  display: inline-block;
```
rogatty: We use tabs.

rogatty:
Moved to https://github.com/Wikia/design-system/pull/36

PR #36, opened 2016-08-23
XW-1903 | Add dropdown styles

rogatty:
https://wikia-inc.atlassian.net/browse/XW-1903
https://github.com/Wikia/design-system-reference-page/pull/62


tomasznapieralski:
LGTM

components/_dropdowns.scss
```diff
@@ -0,0 +1,38 @@
...
+
+.wds-dropdown-toggle {
+	position: relative;
+	// dropdown content has shadow that should be hidden below the toggle
+	z-index: $dropdown-content-z-index + 1;
+}
+
+.wds-dropdown-content {
+	background-color: $color-white;
+	box-shadow: 0 0 10px 0 rgba(0, 0, 0, .2);
```
kvas-damian: I'd feel better if we have have variable for black color ;)

kvas-damian:
One minor comment, besides that it LGTM.

components/_dropdowns.scss
rogatty: Let's address it later when we decide what to do with `$color-black` which isn't black.

PR #37, opened 2016-08-23
XW-1687 | static logged in global nav

Warkot:
https://wikia-inc.atlassian.net/browse/XW-1687
https://github.com/Wikia/design-system-reference-page/pull/63

@Wikia/x-wing


rogatty:
LGTM

components/_global-navigation.scss
```diff
@@ -93,21 +93,18 @@ $notification-counter-border-width: 1px;
...
 		fill: $color-dark-blue-gray;
 		height: 10px;
-		vertical-align: middle;
 		width: 10px;
 	}
 
 	&__content-bar {
 		border-top: $border-top-width solid $color-slate-gray;
 		color: $color-dark-blue-gray;
-		display: inline-block;
```
kvas-damian: I had a feeling it was needed here to make ellipsis work on longer translations.

components/_global-navigation.scss
```diff
@@ -57,10 +59,52 @@ $nav-width: 1024px;
...
+	&__notification-counter-wrapper {
+		background-color: $color-tv;
+		border: 1px solid $color-white;
+		border-radius: $notification-counter-height / 2;
+		color: $color-white;
+		font-size: $typescale-size-minus-3;
+		height: $notification-counter-height - 2 * $notification-counter-border-width;
+		left: 6px;
+		line-height: $notification-counter-height - 2 * $notification-counter-border-width;
+		min-width: $notification-counter-height - 2 * $notification-counter-border-width;
```
kvas-damian: while `$notification-counter-height` is used for `width` calculations I'd name it differently: eg `$notification-counter-size`

BTW: we perform the same calculation 3 time, maybe it's time for a variable or changing `box-sizing` attribute for this component.

components/_global-navigation.scss
```diff
@@ -57,10 +59,52 @@ $nav-width: 1024px;
...
+		padding: 0 4px;
+	}
+
+	&__icon-bell {
+		fill: $color-dark-blue-gray;
+		height: 18px;
+		width: 18px;
+	}
+
+	&__icon-dropdown {
```
kvas-damian: I think this icon should be part of `dropdown` component and ready to be reused. Right now class name suggest it's global nav specific.

components/_global-navigation.scss
```diff
@@ -75,6 +119,12 @@ $nav-width: 1024px;
 			width: 80px;
 		}
 
+		&.wds-has-multiple-subitems {
```
kvas-damian: Do we need a separate modifier for it? Maybe we could merge those styles into `content-bar`?

components/_global-navigation.scss
rogatty: We can use `.wds-dropdown-toggle-chevron` instead of this one after https://github.com/Wikia/design-system/pull/36 is merged. Do you think we should add these properties there?
```
fill: $color-dark-blue-gray;
height: 10px;
width: 10px;
```

components/_global-navigation.scss
Warkot: It works without it. At least on my Chrome.

components/_global-navigation.scss
Warkot: +1 to changing it after merge and adding these styles there (can be done after merge too).

components/_global-navigation.scss
Warkot: My first idea was the same as yours, but I couldn't center the text inside the `fixed-width` containers. If you have some tips or just know how to handle it—it would be appreciated.

components/_global-navigation.scss
```diff
@@ -57,10 +59,52 @@ $nav-width: 1024px;
 		}
 	}
 
+	&__icon-user {
```
hakubo: should that become .wds-avatar.wds-is-circle?

components/_global-navigation.scss
```diff
@@ -57,10 +59,52 @@ $nav-width: 1024px;
...
+		min-width: $notification-counter-height - 2 * $notification-counter-border-width;
+		position: absolute;
+		top: 7px;
+	}
+
+	&__notification-counter {
+		padding: 0 4px;
+	}
+
+	&__icon-bell {
```
hakubo: should .wds-global-navigation__icon-bell and .wds-global-navigation__icon-dropdown
become
.wds-icon-bell and .wds-icon-dropdown ?

components/_global-navigation.scss
```diff
@@ -57,10 +59,52 @@ $nav-width: 1024px;
...
+		margin-right: 2px;
+		width: 28px;
+	}
+
+	&__notifications {
+		position: relative;
+	}
+
+	&__notification-counter-wrapper {
+		background-color: $color-tv;
```
hakubo: it sounds like a coincidence that background color of counter-wrapper has tv color
what if tv color change? should this change as well?

components/_global-navigation.scss
Warkot: What do you think about such solution?
```
$notification-counter-height: 18px;
$notification-counter-border-width: 1px;
$notification-counter-size: $notification-counter-height - 2 * $notification-counter-border-width;
…
&__notification-counter-wrapper {
	…
	border-radius: $notification-counter-height / 2;
	color: $color-white;
	font-size: $typescale-size-minus-3;
	height: $notification-counter-size;
	left: 6px;
	line-height: $notification-counter-size;
	min-width: $notification-counter-size;
	…
}
```

components/_global-navigation.scss
Warkot: This should be this one I believe (from `colors.scss`):
```
// ## Fandom Corporate Colors
$color-link: #00b7e0;
```
Or should I define one more color? We used to have all color names defined and then used them for feature colors:
```
$color-aqua-blue: #00b7e0;
$color-tv: $color-aqua-blue;
```
Maybe we should go back to this?

components/_global-navigation.scss
Warkot: Do you suggest making it a separate component? That makes sense—avatar is also used e.g. on discussions in posts.

components/_global-navigation.scss
Warkot: Dropdown will be changed after merge with the dropdown branch https://github.com/Wikia/design-system/pull/37#discussion_r76014586.

Where would you define the bell icon? Should it be a separate component? Will we do something like that with all the icons?

components/_global-navigation.scss
kvas-damian: What do you think about using `box-sizing: border-box` for this element? It'd save us these calculations and perform them in the browser.

components/_global-navigation.scss
kvas-damian: I hope we won't forget about it.

components/_global-navigation.scss
kvas-damian: It's very likely we'll drop `fixed-width` completely

components/_global-navigation.scss
rogatty: Let's move it to dropdowns, merge them first and pull new master here. That way we won't forget.

components/_global-navigation.scss
kvas-damian: I don't think we should make bell as a separate component at this point. I don't see any additional use-case currently, so I'd suggest keeping it as it is. We can always reiterate when we find it required.

components/_global-navigation.scss
kvas-damian: What's wrong with using `$color-link`?

components/_global-navigation.scss
Warkot: Will it be/is it done in the base branch (XW-1618)? If these will be dropped I might overwrite `display: flex;` for them back to `block` or `inline-block`.

components/_global-navigation.scss
Warkot: Unfortunately `line-height` still needs these 2px less. 😞

rogatty:
LGTM

components/_global-navigation.scss
```diff
@@ -60,6 +94,20 @@ $nav-height: 49px;
 		text-align: center;
 		text-decoration: none;
 
+		&.wds-has-fixed-width {
```
kvas-damian: I believe we already removed this class ;)

components/_global-navigation.scss
```diff
@@ -47,25 +55,101 @@ $nav-height: 49px;
...
 		display: flex;
+		justify-content: center;
+	}
+
+	&__dropdown {
+		color: $color-dark-blue-gray;
+		width: 190px;
+		z-index: -1;
+
+		.wds-button {
```
Warkot: Should we use inheritance here and style based on context? Or should we create separate classes like `wds-global-navigation__account-navigation-button`? Or maybe add a modifier to the base definition of the (in this case) button?

components/_avatar.scss
```diff
@@ -0,0 +1,9 @@
+.wds-avatar {
+	border: solid 2px currentColor;
+	height: 28px;
+	width: 28px;
```
kvas-damian: Sizes for squares could be extracted into a variable.

components/_global-navigation.scss
```diff
@@ -1,16 +1,24 @@
 $border-top-width: 6px;
 $nav-height: 49px;
+$nav-background-color: $color-white;
+$nav-button-active-color: darken($color-link, 20%);
```
kvas-damian: Not used variable.

components/_global-navigation.scss
```diff
@@ -1,16 +1,24 @@
...
+
+$notification-counter-size: 18px;
+$notification-counter-border-width: 1px;
```
kvas-damian: Not used variables.

components/_global-navigation.scss
```diff
@@ -114,6 +154,53 @@ $nav-height: 49px;
...
+	}
+
+	&__user-icon {
+		height: 16px;
+		margin-right: 5px;
+		width: 16px;
+	}
+
+	&__notifications {
+		flex: initial;
```
kvas-damian: `initial` is not supported by IE

components/index.scss
```diff
@@ -1,3 +1,6 @@
+@import '../bower_components/flexboxgrid/dist/flexboxgrid';
```
kvas-damian: It's not part of components anymore. Probably we didn't resolve merge conflicts correctly.

hakubo:
lgtm

components/_dropdowns.scss
```diff
@@ -25,7 +29,9 @@ $dropdown-content-z-index: 1;
 	background-color: $color-white;
 	box-shadow: 0 0 10px 0 rgba($color-black, .2);
 	display: none;
+	left: 0;
 	position: absolute;
+	top: calc(100% + 1px);
```
kvas-damian: Why is it `1px`? Is it related to border?

Warkot: Yup, bottom nov "border".

kvas-damian: isn't nav specific style?
I'm not sure if it should be part of dropdown component.

PR #38, opened 2016-08-23
XW-1619 | add list component styles

tomasznapieralski:
https://github.com/Wikia/design-system-reference-page/pull/64

Duplicate for https://github.com/Wikia/design-system/pull/33, but target branch is `master` here

### DIFF
--- a/components/_lists.scss
+++ b/components/_lists.scss
@@ -26,13 +26,11 @@
 		}
 	}
 
-	&.wds-has-lines-between {
-		> li {
-			border-bottom: solid 1px $color-light-c3;
+	&.wds-has-lines-between > li {
+		border-bottom: solid 1px $color-light-c3;
 
-			&:last-child {
-				border: 0;
-			}
+		&:last-child {
+			border: 0;
 		}
 	}
 }
--- a/components/_lists.scss
+++ b/components/_lists.scss
@@ -1,19 +1,19 @@
 .wds-list {
-  display: block;
-  list-style-type: none;
-  padding: 0;
-  margin: 18px;
+	display: block;
+	list-style-type: none;
+	padding: 0;
+	margin: 18px;
 }
 
 .wds-list-item {
-  display: flex;
-  flex-direction: column;
-  font-size: $typescale-size-minus-1;
-  justify-content: center;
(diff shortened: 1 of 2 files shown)


rogatty:
LGTM

components/_lists.scss
```diff
@@ -0,0 +1,19 @@
...
+	margin: 18px;
+	padding: 0;
+}
+
+.wds-list-item {
+	display: flex;
+	flex-direction: column;
+	font-size: $typescale-size-minus-1;
+	justify-content: center;
+	min-height: 36px;
```
kvas-damian: Should we use `min-height` or `padding` here? What would happen if there is a multiline item?

kvas-damian:
What about list with separators? Weren't we planning to implement it as part of this ticket?

components/_lists.scss
```diff
@@ -0,0 +1,19 @@
+.wds-list {
+	display: block;
+	list-style-type: none;
+	margin: 18px;
+	padding: 0;
+}
+
+.wds-list-item {
+	display: flex;
```
Warkot: Using flex for an item makes all tags in the item to be displayed as a separate flex item. Putting even an inline element such as `<a>` inside the list item will break the line.

components/_lists.scss
```diff
@@ -0,0 +1,32 @@
...
+
+	> li {
+		box-sizing: border-box;
+		font-size: $typescale-size-minus-1;
+		line-height: 1em;
+		padding: 11px 0;
+	}
+
+	&.wds-has-big-items {
+		margin: 2px 18px;
```
rogatty: Or:
```
margin: {
  bottom: 2px;
  top: 2px;
}

File: components/_lists.scss [hunk 4]
```diff
@@ -0,0 +1,32 @@
...
+
+		> li {
+			font-size: $typescale-size-base;
+			padding: 16px 0;
```
rogatty: Or:
```
padding: {
  bottom: 16px;
  top: 16px;
}
```

rogatty:
Minors but overall LGTM

Warkot:
LGTM

PR #40, opened 2016-08-25
XW-1618 | Static global nav on flexgrid

kvas-damian:
https://wikia-inc.atlassian.net/browse/XW-1618

Ping @Wikia/x-wing


components/_global-navigation.scss
```diff
@@ -0,0 +1,122 @@
...
+			height: 15px;
+			width: 42px;
+		}
+
+		&__powered-by {
+			padding: 2px 5px 0;
+		}
+	}
+
+	&__middle-column {
```
rogatty:
```
text-align: center
vertical-align: middle
```
I'd use `&__center-column`

dist/styles.scss
```diff
@@ -1,2 +1,3 @@
 @import '../base/index';
 @import '../components/index';
+@import '../bower_components/flexboxgrid/dist/flexboxgrid.min.css';
```
rogatty: It's already imported in `components/index`. Does it make sense to do it again?

components/_global-navigation.scss
kvas-damian: Discussed offline - middle makes sense here ;)

base/index.scss
```diff
@@ -1,3 +1,4 @@
+@import '../bower_components/flexboxgrid/dist/flexboxgrid';
```
tomasznapieralski: i would move this import to `dist/styles.scss`, because `index` should contain partials in this directory not from outside

tomasznapieralski:
One minor, LGTM

PR #41, opened 2016-08-26
XW-1618 | Fix global nav for 1024px screens + tweak logo + change buttons color

kvas-damian:
@Wikia/x-wing

### DIFF
--- a/components/_global-navigation.scss
+++ b/components/_global-navigation.scss
@@ -116,9 +116,4 @@ $nav-height: 49px;
 	&__account-navigation {
 		flex: 0 1 50%;
 	}
-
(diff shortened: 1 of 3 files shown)


components/_global-navigation.scss
```diff
@@ -10,10 +10,10 @@ $nav-height: 49px;
 	&__content-bar {
 		height: $nav-height;
 		margin: 0 auto;
-		width: 1024px;
+		width: 1004px;
```
rogatty: I wonder if we should have `width: 1024px - $scrollbar-width` just to make the logic behind that number obvious.

hakubo:
lgtm

Warkot:
LGTM

PR #42, opened 2016-08-26
XW-1620 | Search in global nav

kvas-damian:
https://wikia-inc.atlassian.net/browse/XW-1620
https://github.com/Wikia/design-system-reference-page/pull/68
Ping @Wikia/x-wing


components/_global-navigation.scss
```diff
@@ -143,9 +143,78 @@ $notification-counter-size: 18px;
...
+		}
+
+		&__label {
+			align-items: center;
+			display: flex;
+			flex: 1;
+		}
+
+		&__input {
+			border: none;
```
tomasznapieralski: how about 0?

Warkot: https://github.com/Wikia/guidelines/blob/master/SCSS/CodingConventions.md#borderzero

components/_global-navigation.scss
```diff
@@ -143,9 +143,78 @@ $notification-counter-size: 18px;
...
+					margin-right: 15px;
+				}
+
+				&__input {
+					&::placeholder {
+						color: $color-light-gray;
+					}
+				}
+
+				&__submit, &__close {
```
tomasznapieralski: break this on 2 lines

tomasznapieralski:
Minors, minors, LGTM

components/_global-navigation.scss
Warkot: I would break the line after comma, for stronger visual separation of two selectors.

components/_global-navigation.scss
```diff
@@ -143,9 +143,79 @@ $notification-counter-size: 18px;
...
+			}
+		}
+
+		&.wds-is-active {
+			.wds-global-navigation__search {
+				&__input-wrapper {
+					margin-right: 15px;
+				}
+
+				&__input {
```
Warkot: Maybe just`&__input::placeholder {…`?

kvas-damian: It doesn't really matter to me. Which way do you prefer?

rogatty:
We should retarget it onto the master branch.

components/_global-navigation.scss
rogatty: As long as there are no more rules than the current ones it makes sense to keep it simple and without the nesting.

components/_global-navigation.scss
```diff
@@ -215,3 +266,26 @@ $notification-counter-size: 18px;
 		}
 	}
 }
+
+
+
+.wds-search-is-active {
+	.wds-global-navigation__search {
```
rogatty: I think it should be where the other styles for this class are, like it was before the last commit. It will be harder to find as it is now.

Warkot:
LGTM

PR #52, opened 2016-09-05
XW-1928 | Support for search input behaviour

Warkot:
https://github.com/Wikia/app/pull/11268

@Wikia/x-wing


Grunny:
:+1:

components/_global-navigation.scss
```diff
@@ -171,7 +170,7 @@ $notification-counter-size: 18px;
 		// has `width` set in `%` and is inside `flexbox`.
 		min-width: 0;
 		padding: 3px 4px;
-		width: 100%;
+		width: 0;
```
kvas-damian: This element has padding that we don't hide on tablet breakpoint - it ends up with more space between search icon and user menu.

kvas-damian:
Are we going to change cursor to `pointer` for search label as well here?

components/_global-navigation.scss
Warkot: I first wanted to check if it actually works. And it seems it does. Thanks for pointing this out. I will fix it. 👍

Warkot:
IMO, yes, we should change it here as it concerns the same issue/element.

Warkot:
VRT mobile regression looks really weird. Like all screens besides the initial one have no background. http://jenkins.wikia-prod:8080/job/design-system-reference-page-pr-check/403/artifact/reference-page/visuals/reference-page/reference-page.chrome-mobile.regression.png
Any idea why come? 😕

components/_global-navigation.scss
```diff
@@ -300,13 +308,18 @@ $notification-counter-size: 18px;
...
-		&__search-input {
-			display: block;
-		}
-
 		&__search-input-wrapper {
 			border-bottom-width: 1px;
 		}
+
+		&__search-label-icon:hover {
+			color: $color-dark-blue-gray;
```
rogatty: Can we do it in a way that won't need change in multiple places when we decide to change this default color? E.g. use a variable for all navigation default colors.

components/_global-navigation.scss
```diff
@@ -300,13 +308,18 @@ $notification-counter-size: 18px;
...
+		}
+
+		&__search-input {
+			padding: $search-input-padding;
+			width: 100%;
```
rogatty: We have this in 3 places, can we keep it DRY?

rogatty:
One thing and will LGTM

tomasznapieralski:
LGTM

Warkot:
@rogatty is it DRY enough? https://github.com/Wikia/design-system/pull/52/commits/6c4290239bee149c65174f7090894bbbe9d446cd

rogatty:
LGTM

components/_global-navigation.scss
```diff
@@ -3,8 +3,29 @@ $nav-height: 49px;
 $nav-background-color: $color-white;
 
 $logo-width: 230px;
+$search-input-padding: 3px 4px;
 $notification-counter-size: 18px;
 
+@mixin visible-search-input {
```
kvas-damian: I don't like the idea of including mixins multiple times for styling the same elements.
Do you think it's possible to simplify that code somehow?

reference-page/config/custom-styles.css
```diff
@@ -1,3 +1,7 @@
+body {
+	background-color: #fff;
```
kvas-damian: What happened that we need to add this style?

rogatty: Something strange with VRT / WebdriverCSS / Selenium. See http://jenkins.wikia-prod:8080/job/design-system-reference-page-pr-check/403/artifact/reference-page/visuals/reference-page/reference-page.chrome-mobile.regression.png

components/_global-navigation.scss
Warkot: Let's think about it after the release. 👍

PR #54, opened 2016-09-06
(XW-1969) Build compiled CSS file to distribute

Grunny:


gulpfile.js
```diff
@@ -92,6 +92,17 @@ gulp.task('svg-individual', function () {
 gulp.task('svg', ['svg-sprite', 'svg-individual']);
 
 
+// BUILD DESIGN SYSTEM STYLES
+gulp.task('build-styles', function () {
+	return gulp.src('dist/styles.scss')
+		.pipe(scss())
+		.pipe(autoprefixer({
+			browsers: ['last 3 versions'],
```
kvas-damian: I see mercury is using different config for supported browsers: https://github.com/Wikia/mercury/blob/4eb062496b96876656dde7c29147582638a10727/front/auth/gulpfile.js#L45
Do we need to do something about it?

gulpfile.js
```diff
@@ -92,6 +92,17 @@ gulp.task('svg-individual', function () {
...
+			cascade: false
+		}))
+		.pipe(gulp.dest('dist/css'));
```
kvas-damian: Is it possible to minimize output file?

Grunny: Done in 4a7ef490cbd4221ff8dc95158075586dbd23f574.

gulpfile.js
Grunny: I'm not sure. This just matches what we're doing for reference page. Someone with better knowledge on why the config options were chosen for both RP and Mercury would be able to make a call.

kvas-damian:
What happened to VRT that we had to update baseline in this PR?
I don't what we changed on a Reference Page.

Grunny:
I'm not sure, I didn't do that. Was it cherry-picked? Nothing should have changed on the reference page. I can rebase against master and get rid of that commit to clean it up.

kvas-damian:
LGTM

rogatty:
I've updated the VRT baseline. Something changed with VM / Selenium / Chrome or whatever and currently VRT fails for mobile version on all branches.

rogatty:
LGTM

PR #62, opened 2016-09-14
Update README.md

lizlux:
Adding info about the design system API so that contributors to this repo can be aware of the larger system surrounding it. 

@Wikia/west-wing @Wikia/x-wing


tomasznapieralski:
Great idea @lizlux those links would be very useful for people who just started using Wikia/design-system :)

PR #63, opened 2016-09-15
XW-2018 | animations for global navigation

Warkot:
* https://wikia-inc.atlassian.net/browse/XW-2018
* https://github.com/Wikia/app/pull/11344
* https://github.com/Wikia/mercury/pull/2791

@Wikia/x-wing


components/_global-navigation.scss
```diff
@@ -250,10 +278,14 @@ $notification-counter-size: 18px;
 	}
 
 	&__search-close {
-		display: none;
+		display: block;
```
hakubo: do we still need that?

components/_global-navigation.scss
```diff
@@ -343,36 +380,51 @@ $notification-counter-size: 18px;
 	}
 
 	&.wds-search-is-active .wds-global-navigation {
-		@include visible-search-input;
+		//@include visible-search-input;
```
hakubo: should that be removed?

components/_global-navigation.scss
Warkot: No, we just need these for single SVGs inside buttons. 👍

components/_global-navigation.scss
weronika: right, it's useless

components/_global-navigation.scss
```diff
@@ -153,6 +157,18 @@ $notification-counter-size: 18px;
...
+	&__links-and-search {
+		display: flex;
+		flex: 4 1 auto;
+	}
+
+	&__link,
+	&__wikis-menu {
+		max-width: 100%;
+		transition-duration: $search-transition-duration;
+		transition-property: color, flex, font-size, max-width, padding;
```
tomasznapieralski: you can group those `transition` properties

components/_global-navigation.scss
```diff
@@ -193,12 +209,19 @@ $notification-counter-size: 18px;
 		flex: 0 1 auto;
 	}
 
+	&__search {
+		flex: 0 1 auto;
+		transition-duration: $search-transition-duration;
+		transition-property: flex;
```
tomasznapieralski: you can group those `transition` properties

components/_global-navigation.scss
```diff
@@ -221,19 +246,21 @@ $notification-counter-size: 18px;
...
-		width: 1px;
+		left: 0;
+		max-width: 100%;
+		padding: $search-input-padding;
+		padding-left: 20px;
+		position: absolute;
+		right: 0;
+		top: 3px;
+		transition-duration: $search-transition-duration;
+		transition-property: width, padding;
```
tomasznapieralski: you can group those `transition` properties

components/_global-navigation.scss
```diff
@@ -250,10 +277,13 @@ $notification-counter-size: 18px;
...
 	&__search-close {
-		display: none;
 		// Safari fix for bleeding search suggestions shadow
 		margin: 0;
-		padding: 3px;
+		max-width: 0;
+		opacity: 0;
+		padding: 0;
+		transition-duration: $search-buttons-transition-duration;
+		transition-property: max-width, opacity, padding;
```
tomasznapieralski: you can group those `transition` properties

components/_global-navigation.scss
```diff
@@ -263,8 +293,13 @@ $notification-counter-size: 18px;
...
 
 	&__search-submit {
-		display: none;
-		padding: 6px 9px;
+		border-width: 0;
+		max-width: 0;
+		overflow: hidden;
+		padding: 0;
+		transition-duration: $search-buttons-transition-duration;
+		transition-property: max-width, padding;
```
tomasznapieralski: you can group those `transition` properties

components/_global-navigation.scss
Warkot: When there's multiple properties to be animated in one duration it is much more DRY and readable to write it this way. In this case it would look like below. :dizzy_face:
```
transistion: color $search-transition-duration, flex $search-transition-duration, font-size $search-transition-duration, max-width $search-transition-duration, padding $search-transition-duration;
```

components/_global-navigation.scss
```diff
@@ -193,12 +210,24 @@ $notification-counter-size: 18px;
...
+	}
+
 	&__search-input-wrapper {
-		border: 0 solid $color-slate-gray;
+		border: {
+			bottom: 1px solid transparent;
+			left: 0;
+			right: 0;
+			top: 0;
+		};
```
kvas-damian: What do you think about 
```
border: 0 solid transparent;
border-bottom-width: 1px;
```
?

components/_global-navigation.scss
```diff
@@ -153,6 +158,18 @@ $notification-counter-size: 18px;
...
+	&__links-and-search {
+		display: flex;
+		flex: 4 1 auto;
+	}
+
+	&__link,
+	&__wikis-menu {
+		max-width: 100%;
+		transition-duration: $search-transition-duration;
+		transition-property: color, flex, font-size, max-width, padding;
```
kvas-damian: Do we need to animate both `flex` and `max-width`?

components/_global-navigation.scss
```diff
@@ -193,12 +210,24 @@ $notification-counter-size: 18px;
...
+			bottom: 1px solid transparent;
+			left: 0;
+			right: 0;
+			top: 0;
+		};
 		display: flex;
 		// Safari doesn't set default flex values if we use `flex: 1;`
 		// it requires us to put these values explicitly.
 		flex: 1 1 auto;
+		min-height: 28px;
```
kvas-damian: why do we need to set `min-height` on a wrapper?

components/_global-navigation.scss
```diff
@@ -221,19 +252,23 @@ $notification-counter-size: 18px;
...
+		height: 100%;
+		left: 0;
+		max-width: 100%;
+		padding: $search-input-padding;
+		padding-left: 20px;
+		position: absolute;
+		right: 0;
+		top: 0;
+		transition-duration: $search-transition-duration;
+		transition-property: width, padding;
```
kvas-damian: Don't we need initial `width` value for animating this property?

components/_global-navigation.scss
```diff
@@ -338,41 +386,59 @@ $notification-counter-size: 18px;
...
 		}
 	}
 
 	&.wds-search-is-active .wds-global-navigation {
-		@include visible-search-input;
 		@include search-label-icon-default-hover-color;
 
 		&__link,
 		&__wikis-menu {
+			color: rgba($color-dark-blue-gray, 0);
```
kvas-damian: cant we simply use `transparent` or `white` here?

components/_global-navigation.scss
```diff
@@ -221,19 +252,23 @@ $notification-counter-size: 18px;
...
-		padding: 0;
-		// Safari doesn't trigger focus on input if it has `width 0;`
-		width: 1px;
+		// iPad Safari needs the height below to be set
+		height: 100%;
+		left: 0;
+		max-width: 100%;
+		padding: $search-input-padding;
+		padding-left: 20px;
+		position: absolute;
```
kvas-damian: In which case we need to use `position: absolute`?

components/_global-navigation.scss
Warkot: We could set just `height`, but it's needed to keep the same size of the wrapper whether it's closed, open with or without buttons. But I'll give it another check.

components/_global-navigation.scss
Warkot: This way we set unnecessary values for all other borders then bottom one. And we immediately overwrite a value that's needed only for the bottom border. I'm not (yet 😉 ) convinced to your proposed solution.

components/_global-navigation.scss
Warkot: Let's check it. 👍

components/_global-navigation.scss
weronika: I like Damian's suggestion, this notation is more clear. Another alternative is:

```
border: solid transparent;
border-width: 0 0 1px 0;
```

which I like most ;)

kvas-damian: On the other hand shorthand style is easier to be overwritten.

components/_global-navigation.scss
weronika: We need it on small breakpoint - if it's not absolute, it breaks the animation, because it pushes its own container. We can't animate its width either, 'cause it would be totally separate animation and it would not run synchronuosly with the main animation (I've tested it).

components/_global-navigation.scss
Warkot: Nope, removed. 😃

components/_global-navigation.scss
weronika: it works without it, but we'll add it, just for readiness

weronika: I've checked it again and it appeared that we don't need animating width any more. so we'll remove it.

components/_global-navigation.scss
Warkot: Yup, it's needed. I'll still change it to `height` to be more explicit.

PR #65, opened 2016-09-19
XW-2003 | update icons in Global Navigation

Warkot:
ℹ️  We have to rethink how we're implementing the icons in Design System in broader view. This PR is blocked by that. It will most probably get closed and redone in the new way.

https://wikia-inc.atlassian.net/browse/XW-2003
https://github.com/Wikia/app/pull/11381
https://github.com/Wikia/mercury/pull/2798

@Wikia/x-wing


assets/icons/user.svg
```diff
@@ -1,3 +1,3 @@
-<svg width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
-  <path d="M12 14c3.309 0 6-2.691 6-6V6c0-3.309-2.691-6-6-6S6 2.691 6 6v2c0 3.309 2.691 6 6 6zm5 2H7c-3.86 0-7 3.14-7 7a1 1 0 0 0 1 1h22a1 1 0 0 0 1-1c0-3.86-3.14-7-7-7z" fill-rule="evenodd"/>
+<svg width="18" height="18" viewBox="0 0 18 18" xmlns="http://www.w3.org/2000/svg">
```
hakubo: we should have consistent sizes of icons. 
As discussed offline let's meet this week and talk whether we really need different versions for different sizes or can we send fewer bytes to browsers

reference-page/components/global-navigation.md
```diff
@@ -147,7 +147,7 @@
 			<div class="wds-global-navigation__notifications-menu wds-dropdown">
 				<div class="wds-dropdown__toggle wds-global-navigation__dropdown-toggle">
 					<div class="wds-global-navigation__notifications-menu-counter">3</div>
-					<svg class="wds-icon wds-icon-small" alt="notifications">
```
hakubo: I'd keep this class here
if we'll go back to 24x24 as a default we would have to re-add it here and in all apps that implement it

tomasznapieralski:
Let's do not keep it open, if we will work on it later we can reopen this

PR #71, opened 2016-09-26
Remove redundant badges

rogatty:
Those files are now available in `dist/svg/wds-avatar-*`. Removing the duplicates.

/cc @Wikia/x-wing @Wikia/west-wing


kvas-damian:
Should we have this cleanup automated?

rogatty:
@kvas-damian I've tried and failed. Our build pipeline is very brittle and doesn't like changes. Let's fix it when rebuilding the RP.

PR #77, opened 2016-10-10
FAN-985 Remove unneeded transforms from SVGs that break in Chrome

mklucsarits:
The way in which we're using SVG sprites on Fandom Stories exposed a bug in Chrome. I don't entirely understand the issue, except that we had two SVGs not displaying properly as a result. However, the two files shared a pattern: They had both had group (`<g>`) tags containing masks separate from the main paths. This allowed me to propose a fix. I asked Derek H. to edit the files in Sketch, and to move the paths in separate groups into the main layer with the other paths. This resolved the issue.

This bug was not occurring on the Mediawiki stack. The difference being that on Wikia/app, each entire SVG is included inside the `<svg>` tag, whereas on Fandom stories we use `<use>` like Mercury. The separate transform groups in the problematic symbols also had their own `<use>` tags linking to paths in the combined/inlined `<defs>`. I would guess these multiple `<use>` links play a part in the underlying issue, although I'm not sure why.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "design-system",
-  "version": "0.14.4",
+  "version": "0.14.5”,
```
rogatty: This is causing build to fail:
```
npm ERR! Failed to parse json
npm ERR! Unexpected token '\n' at 3:24
npm ERR!   "version": "0.14.5â€�,
```
http://jenkins.wikia-prod:8080/job/design-system-reference-page-pr-check/705/console

rogatty: After you fix it please check the VRT results to make sure there is no visual difference.

assets/company/logo-fandom-powered-by-wikia-two-lines.svg
```diff
@@ -1,33 +1,45 @@
...
-    <g transform="translate(127.537 10.105)">
-      <mask id="logo-fandom-powered-by-wikia-two-lines-mask-1" fill="#fff">
-        <use xlink:href="#logo-fandom-powered-by-wikia-two-lines-path-1"/>
-      </mask>
-      <path d="M24.028 2.343C22.668.821 20.748.016 18.476.016c-2.314 0-4.21 1.106-5.377 3.125C11.933 1.122 10.036.016 7.723.016 5.451.016 3.53.82 2.17 2.343.902 3.763.204 5.753.204 7.947v9.086H4.49V7.947c0-2.454 1.178-3.862 3.233-3.862s3.233 1.408 3.233 3.862v9.086h4.286V7.947c0-2.454 1.179-3.862 3.234-3.862 2.054 0 3.233 1.408 3.233 3.862v9.086h4.286V7.947c0-2.193-.699-4.184-1.967-5.604" fill="#FFF" mask="url(#logo-fandom-powered-by-wikia-two-lines-mask-1)"/>
+<svg width="184px" height="44px" viewBox="0 0 184 44" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
rogatty: Please use width and height without `px` to be consistent with the rest of the files.
Also, run `npm run svgo` to optimize the source files. There is a lot of ids, transforms, paths and other unnecessary cruft here.

package.json
mklucsarits: There are visual differences. The reference page is completely different even though all I did was update the SVGs.

assets/company/store-appstore.svg
```diff
@@ -1,13 +1,37 @@
...
-  </g>
+<svg width="119px" height="35px" viewBox="0 0 119 35" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
mklucsarits: That's correct. The problem with the Google Play image was the gradient was not displaying, due to id conflicts in the inlined SVG. https://github.com/Wikia/upstream/pull/733 resolves that. The Apple store image then had the problems described in this PR. So they are two separate issues.

package.json
mklucsarits: That's what happens when you try to quickly edit the version number in TextEdit... :(

mklucsarits:
@rogatty Can you take another look at this please?

mklucsarits:
Thanks, @rogatty !

mklucsarits:
For some reason the SVGs in /dist still have transforms when the originals do not. Why would that be, if they are only minified with gulp? As a result, those SVGs are still broken after this PR has been merged.

PR #87, opened 2016-10-21
XW-2232 | add slot to global nav that alignes itself to right and disappears wh…

hakubo:
…en screen is narrower than 1280px


style-guide/styles/wds-components/_global-navigation.scss
```diff
@@ -377,6 +378,21 @@ $hover-transition-duration: 300ms !default;
...
+	&__right-slot {
+		max-width: 100px;
+		display: none;
+		height: 100%;
+		position: absolute;
+		right: 0;
+		top: 0;
+		border-left: 1px solid $color-blue-gray;
+
+		@media (min-width: 1280px) {
```
rogatty: Where is this number coming from?

hakubo: arbitrary but very well chosen number - it just fits only when your screen is 1280px+,
I can add a comment about it

lib/wds-components/app/templates/components/wds-global-navigation.hbs
```diff
@@ -36,3 +36,8 @@
 		</a>
 	</div>
 </div>
+<div class="wds-global-navigation__right-slot">
+	<a href="" class="wds-global-navigation__right-slot-link">
+		<img class="wds-global-navigation__right-slot-image" src="{{concat rootURL 'images/test.png'}}" alt="test">
```
rogatty: Let's add it to `/app/models/global-navigation.json` and use `model` here.

PR #116, opened 2017-03-08
fixing border style override

lizlux:
The `border: 1px solid;` property was overriding the border color set in the mixin. This will let border color, width, and style be set independently of one another. Should fix the 1px height difference above and below primary buttons. 
screen shot 2017-03-08 at 2 16 23 pm

@Wikia/x-wing @vforge


rogatty:
Please take a look at http://fandomdesignsystem.com/#/overview/contributing and:
* run `npm run build` to update the `dist` directory
* after this is merged to master, run `npm run release-fix` which will create the release and bump the package.json version
* update changelog on https://github.com/Wikia/design-system/releases

The change itself LGTM.

lizlux:
Thanks @rogatty! Project is built.

PR #117, opened 2017-03-15
FAN-1743 Add additional SVG Icons

JoshuaRogan:
This PR adds three additional icons for use on Fan Contributor editing tools.


docs/svg/wds-icons-more-horizontal.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
```
kvas-damian: isn't it enough to use https://github.com/Wikia/design-system/blob/master/docs/svg/wds-icons-more.svg and rotate it in css?

PR #122, opened 2017-04-05
IRIS-3991 | Styles for Sign Out

slayful:


dist/scss/wds-components/_buttons.scss
```diff
@@ -98,6 +98,28 @@
 		}
 	}
 
+	&.wds-is-link {
```
kvas-damian: Can't we simply use `wds-is-text` and override styles that are different?

dist/scss/wds-components/_buttons.scss
```diff
@@ -98,6 +98,28 @@
...
+		background: transparent;
+		border: none;
+		color: inherit;
```
kvas-damian: Probably we should have defined one defined color for button.

dist/scss/wds-components/_buttons.scss
```diff
@@ -98,6 +98,28 @@
...
+		cursor: pointer;
+		font: inherit;
+		height: auto;
```
kvas-damian: Do we need to set it? I though it's browsers default.

dist/scss/wds-components/_lists.scss
```diff
@@ -48,5 +48,10 @@
 				padding: $list-item-padding;
 			}
 		}
+
+		.wds-is-link {
```
kvas-damian: I believe this is one time use-case for global navigation. If we really need it let's put it there instead for add more junk into `wds-list`

kvas-damian:
When changing DS please update documentation as well.

PR #124, opened 2017-04-06
XW-3127 | fandom heart icon

latata:
https://wikia-inc.atlassian.net/browse/XW-3127

@Wikia/x-wing


app/templates/route-components/assets.hbs
```diff
@@ -124,6 +124,7 @@
 		{{icon-box name='community' size='small'}}
 		{{icon-box name='eye' size='small'}}
 		{{icon-box name='eye-crossed' size='small'}}
+		{{icon-box name='fandom-heart' size='small'}}
```
kvas-damian: I was never 100% sure about it, but this icon looks like ["Company Assets"](http://wikiadesignsystem.com/#/identity/assets). Shouldn't it be placed there?

latata: Ania approved it in icons

PR #125, opened 2017-04-07
XW-3085 | buttons with icons

latata:
https://wikia-inc.atlassian.net/browse/XW-3085

@Wikia/x-wing


style-guide/styles/wds-components/_buttons.scss
```diff
@@ -98,6 +100,28 @@
 		}
 	}
 
+	&.wds-has-icon {
+		padding-left: $button-padding-horizontal + $wds-icon-small-size + 6px;
+		position: relative;
+
+		&.wds-is-squished {
+			padding-left: $button-squished-padding-horizontal + $wds-icon-tiny-size + 6px;
```
rogatty: Do we assume that using the squished button means using a tiny icon? If so, maybe we should document it somewhere.

latata: added documentation

dist/scss/wds-components/_buttons.scss
```diff
@@ -98,6 +100,28 @@
...
+				left: $button-squished-padding-horizontal;
+				margin-top: -$wds-icon-tiny-size / 2;
+			}
+		}
+
+		.wds-icon {
+			left: $button-padding-horizontal;
+			margin-top: -$wds-icon-small-size / 2;
+			pointer-events: none;
+			position: absolute;
```
kvas-damian: Maybe we could use `disply: flex` for vertical centering?

style-guide/styles/wds-components/_buttons.scss
kvas-damian: Can we make this solution more bullet proof by forcing width and height of the icon?

dist/scss/wds-components/_buttons.scss
latata: we can't because that causes another problems. And it's better to have all buttons same display value - `inline-block`

kvas-damian: what about `inline-flex`? Is it also affected?

PR #130, opened 2017-04-18
IRIS-4218 | Document the Avatar Stack

slayful:


app/templates/route-components/progress-indicators.hbs
```diff
@@ -15,17 +15,17 @@
 		or via JavaScript code.
 	</p>
 	{{#component-demo}}
```
slayful: Removed indents for component-demo contents.

PR #143, opened 2017-05-17
XW-3295 | Hover for dropdowns

rybmat:
https://wikia-inc.atlassian.net/browse/XW-3295

Opening dropdowns on hover

@Wikia/x-wing


rybmat:
will be done together with https://wikia-inc.atlassian.net/browse/XW-3296

PR #153, opened 2017-06-05
XW-3333 | Disabled mixin

JamesHawking:
Added wds-is-disabled as global class.


style-guide/styles/wds-components/_buttons.scss
```diff
@@ -30,8 +30,7 @@
 		cursor: pointer;
 	}
 
-	&:disabled,
-	&.wds-is-disabled {
+	&:disabled {
 		cursor: default;
 		opacity: .5;
```
rogatty: We can use `@include is-disabled();` instead of duplicating these rules.

style-guide/styles/wds-components/_is-disabled.scss
```diff
@@ -0,0 +1,5 @@
+.wds-is-disabled {
+	@include is-disabled();
+}
+
+
```
rogatty: Unnecessary new line.

PR #159, opened 2017-07-13
XW-3630 | active tab indicator theming

latata:
https://wikia-inc.atlassian.net/browse/XW-3630
@Wikia/x-wing


style-guide/styles/wds-mixins/_box-underline.scss
```diff
@@ -0,0 +1,3 @@
+@mixin box-underline($color) {
```
rogatty: How about `wds-box-underline`?

PR #160, opened 2017-07-19
XW-3556 | highlight item in dropdown also with wds-is-selected class

hakubo:


style-guide/styles/wds-components/_dropdowns.scss
```diff
@@ -76,8 +76,10 @@
 						transition-property: background-color, color;
 					}
 
-					&:hover > a {
-						@include wds-dropdown-linked-list-item-hover;
+					&:hover, .wds-is-selected {
+						> a {
+							@include wds-dropdown-linked-list-item-hover;
```
rogatty: The old, currently not-working solution (please clean it up):
https://github.com/Wikia/app/blob/6041aee91d0e29d83e5cfa8cbe1f85b947ce7fad/extensions/wikia/DesignSystem/styles/design-system.scss#L83-L85

It changed only the text color while `wds-dropdown-linked-list-item-hover` changes also the background. Are you sure this is what we want?

PR #161, opened 2017-08-02
IRIS-4622 - update badges

katanka:
https://wikia-inc.atlassian.net/browse/IRIS-4622


rogatty:
The current change updates only icons in the documentation (http://fandomdesignsystem.com/#/components/assets) and will be overridden as soon as someone runs the build. Please put the icons in `/style-guide/assets/` and use `npm run build`.

Make sure that icons are 18x18px, not 18x19px. It's not enough to change the source manually. The asset has to be exported in this size to look good.

Please update element ids to be unique. E.g. instead of `<path id="b"` use `<path id="wds-avatar-badges-global-discussions-moderator-path-1"` (based on https://raw.githubusercontent.com/Wikia/design-system/master/style-guide/assets/wds-company-logo-wikia-org.svg). We don't want duplicated ids on the page.

PR #162, opened 2017-08-03
XW-3616 | extract community header gradient and text color as function/mixin

latata:
We want to reuse code.

@Wikia/x-wing


dist/scss/wds-mixins/_community-header-gradient.scss
```diff
@@ -0,0 +1,5 @@
+@mixin community-header-gradient(
+	$header-background-color, $start-position, $end-position
+) {
+	background: linear-gradient(to right, rgba($header-background-color, .5), rgba($header-background-color, .5)), linear-gradient(to left, rgba($header-background-color, 0) $start-position, $header-background-color $end-position);
```
hakubo: break line

PR #166, opened 2017-08-30
DO NOT MERGE: Grid example

gbenson-wikia:
This is intended to be an example of what a WDS implementation of the page grid might look like, using CSS grids.  I believe it's sufficiently simple that - if others like it - it could be merged into WDS with very little cleanup.

Note that this is only really intended to solve the "how do we express the WDS grid in CSS at the page layout level?", and intentionally doesn't address other spacing concerns in a more general way.

@Wikia/cake 
@vforge

### DIFF
--- a/experiments/grid/index.html
+++ b/experiments/grid/index.html
@@ -5,100 +5,232 @@
 	<meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
 	<title>Grid Experiment</title>
 	<meta name="viewport" content="width=device-width, initial-scale=1">
+	<link rel="stylesheet" href="wikia-design-system-page-layout.css">
 	<link rel="stylesheet" href="wikia-design-system-grid.css">
 	<link rel="stylesheet" href="main.css">
+	<script>
+		function onRemoveBorders() {
+			const elements = document.body.getElementsByTagName("*");
+			for (let i = 0; i < elements.length; i++) {
+				elements[i].style.border = 'none';
+			}
+
+			document.getElementById('removeBorders').style.display = 'none';
+		}
+	</script>
 </head>
-<body class="wds-page-grid">
+<body>
 <header>
-	<h1>I'm a header</h1>
+	<section class="wds-content-well">
+		<section class="wds-content-grid">
+			<div>Header section 1</div>
+			<div>Header section 2</div>
+			<div>Header section 3</div>
+		</section>
+	</section>
 </header>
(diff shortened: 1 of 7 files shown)


lizlux:
We spoke off line about this but I have a few concerns:
* The grid should only be 12 columns total. If you need spacing to the left or right of the grid, create a grid container and adding margins (or flexbox, etc) to center it on the page. The spec (https://zpl.io/29eQPz2) itself isn't actually in-line with what we do in production so this is a conversation we need to have with Derek. 
* The right rail has always been the bane of the grid's existence on wiki pages, since it's a fixed width of 300px. Let's discuss way to work around this constraint.

gbenson-wikia:
A PoC for CSS Custom Properties usage in theming has been added.

In a separate commit, I'll go back and adjust the grid based on conversations with Liz and Derek.

@Wikia/cake 
@vforge

lizlux:
Theming work looks good. What's the long-term plan for this work? Do we want to keep it as an "experiments" section? Or integrate it more fully with the living documentation?

vforge:
@lizlux Since I have a time slot available for the DS work in each sprint, I will be able to plan and implement some things we discussed previously. The first item on the agenda is actually the plan how and what to update - I will work on that this Friday. I'll share my plan after that.

For now, lets it's in the "experiments" only - of course, I encourage any Team to implement their own version (until it's in the DS) - as the Upstream did with Theming and Grid and CAKE with web components.

gbenson-wikia:
The grid experiment has been updated based on feedback from Liz and Derek.

The biggest caveat on this is that the exact set of breakpoints that should be officially supported into the future is presently in flux, so some amount of refactoring would need to happen whenever that settles.

I'm also not super happy with the two-column layout portion, but I'm also having trouble putting my finger on exactly what about it I'm not happy about.  If it were a custom element that accepted exactly two children (instead of a mandated HTML structure & classes), I might like it more.  Anyway, ideas and suggestions on this piece are especially welcome.

@Wikia/cake 
@vforge

gbenson-wikia:
Updated to match the latest grid specification.  The concepts shown by the PoC should be ready for productization into the DS now.
@Wikia/cake 
@vforge

vforge:
Added final PR here: https://github.com/Wikia/design-system/pull/178

vforge:
Released in https://github.com/Wikia/design-system/releases/tag/8.15.0

PR #169, opened 2017-10-03
 XW-4060 | add transparency to banner notification

xkxd:
https://wikia-inc.atlassian.net/browse/XW-4060

@Wikia/x-wing


dist/scss/wds-components/_banner-notification.scss
```diff
@@ -2,6 +2,11 @@
 	background-color: $wds-color-white;
 	color: $wds-color-light-gray;
 	display: flex;
+	transition: opacity 0.4s;
+
+	&--transparent {
```
vforge: Please use `&.wds-is-transparent` - as we're using states (in SMACSS way: https://smacss.com/book/type-state) and not modifiers (in BEM way).
See also https://github.com/Wikia/design-system/blob/master/style-guide/styles/wds-components/_buttons.scss#L41 for examples.

PR #174, opened 2017-12-01
XW-4299 | replace wds-icons-play with updated svg

rybmat:
@Wikia/x-wing


vforge:
@rybmat Previously icons didn't have color and were monochromatic and this one has two defined colors. It slightly changes how the icon can be used and styled. Could we maybe not replace the current one and put this one as separate *asset* (instead of icon)?

style-guide/styles/wds-components/_player.scss
```diff
@@ -0,0 +1,41 @@
+.wds-player-icon-play-tiny {
+	$size: 1.2 * 30px;
```
latata: do we have to multiply by `1.2`? can't we just have `36px`?

style-guide/styles/wds-components/_player.scss
```diff
@@ -0,0 +1,41 @@
...
+
+	height: $size;
```
latata: how about using `square` mixin?

dist/scss/wds-components/_player.scss
```diff
@@ -0,0 +1,41 @@
+.wds-player-icon-play-tiny {
+	$size: 1.2 * 30px;
+
+	height: $size;
```
hakubo: you can use square mixin

hakubo: also SCSS support loop over map so you could generate these clasess
http://thesassway.com/intermediate/if-for-each-while

style-guide/styles/wds-components/_player.scss
rybmat: 👍

style-guide/styles/wds-components/_player.scss
rybmat: @latata I thought it would be easier to understand the logic behind it

PR #179, opened 2018-01-30
MAIN-12770 | Resolve conflict between app and DS buttons

TK-999:
Currently if WDS styles are used on a `<button>`, `<input type="submit" />` or similar element in app, they conflict with 2010 era Oasis styling. By specifying explicit defaults for some settings we will be able to use these DS styles without having to manually override every single time. This helps avoids issues like https://wikia-inc.atlassian.net/browse/MAIN-12770

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .idea/
+*.iml
 *.DS_Store
 *._.DS_Store
 npm-debug.log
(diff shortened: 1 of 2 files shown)


hakubo:
We do shenanigans like this in
https://github.com/Wikia/app/blob/9a49e2e3995c8227576689b8b7e44f070b5bcdfe/extensions/wikia/DesignSystem/styles/design-system.scss

TK-999:
@hakubo cool, thanks! I'll move these changes there.

TK-999:
new PR -> https://github.com/Wikia/app/pull/14607

PR #186, opened 2018-02-27
Link directly to new mediakit url

frankfarmer:
The footer currently links to http://www.wikia.com/mediakit which redirects to fandom.wikia.com/mediakit

Let's take out the middleman and link directly to the page.

### DIFF
--- a/app/models/global-footer.json
+++ b/app/models/global-footer.json
@@ -69,10 +69,10 @@
       "header": { "type": "line-text", "title": { "type": "translatable-text", "key": "Advertise" } }, "links": [
         {
           "type": "link-text", "title": { "type": "translatable-text", "key": "Media Kit" },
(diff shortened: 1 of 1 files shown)


mklucsarits:
@frankfarmer I'm not sure what this JSON is actually used for, but unfortunately it isn't used for the footer. There's a (poorly designed) API which gets links for the global nav and footer from MW. The URL has to be updated in this file: https://github.com/Wikia/app/blob/dev/includes/wikia/models/DesignSystemSharedLinks.class.php

rogatty:
This JSON is used to render demo on http://fandomdesignsystem.com/#/identity/global-footer. You can merge it if you want but it doesn't really matter :)

As documented on http://fandomdesignsystem.com/#/identity/api and as @mklucsarits mentioned, the actual API is implemented in MediaWiki: https://github.com/Wikia/app/pull/14884

PR #190, opened 2018-04-26
XW-4906 | Deploy Reference page to git gh-pages

hakubo:


config/deploy.js
```diff
@@ -2,28 +2,38 @@
...
 
-	if (deployTarget === 'production') {
+	if (deployTarget === 'github') {
 		ENV.build.environment = 'production';
-		// configure other plugins for production deploy target here
+		ENV.plugins.push('git');
+
+		ENV.git = {
+			repo: 'git@github.com:wikia/design-system.git',
+			branch: 'gh-pages',
```
vforge: Instead of `gh-pages` branch you could use `docs/` dir instead.

hakubo: yep, that is how it is right now. 
but that means you have to commit alraedy built reference page

i think workflow
where you do update, commit, merge to master
and then release is more common

thoughts?

hakubo: thanks for reviewing!

vforge: We have the same flow in the react mirror, but we're not using `gh-pages` - it seems simpler that way, as this compiled HTML is always accessible and can be executed from locally (`file://`).

PR #192, opened 2018-04-27
XW-4870 | Split css

rybmat:
https://wikia-inc.atlassian.net/browse/XW-4870
https://github.com/Wikia/mobile-wiki/pull/673

* Split global-footer and buttons css-es to allow mobile-wiki to exclude not used styles
* remove unused obsolete styles from global-footer

@Wikia/x-wing


dist/scss/wds-components/global-footer/_breakpoint-l-down.scss
```diff
@@ -0,0 +1,70 @@
+@import 'variables';
+
+.wds-global-footer {
+	@media #{$wds-breakpoint-l-down} {
```
vforge: Breakpoints in this PR are outdated.

vforge: Yeah, but it's a nice opportunity to replace them:
> NOTICE: The goal is to use those breakpoints in all our products. Please replace the old breakpoints with the new ones whenever and wherever possible.
from http://fandomdesignsystem.com/#/base-styles/wds-breakpoints

PR #197, opened 2018-05-16
XW-4953 | wds-avatar-stack component

rybmat:
https://wikia-inc.atlassian.net/browse/XW-4953

Ember component for avatar-stack

@Wikia/x-wing


app/components/wds-avatar-stack.js
```diff
@@ -0,0 +1,15 @@
+import Component from '@ember/component';
+import {computed} from '@ember/object';
+
+
+export default Component.extend({
+	classNames: ['wds-avatar-stack'],
+	overflow: computed('avatars', function() {
+		const count = this.get('avatars').length;
+
+		return count > 5 ? this.get('avatars').length - 5 : 0;
```
hakubo: this 5 could be a property, so it can be overriden if needed

hakubo: count - 5?

app/components/wds-avatar-stack.js
```diff
@@ -0,0 +1,15 @@
+import Component from '@ember/component';
+import {computed} from '@ember/object';
+
+
+export default Component.extend({
+	classNames: ['wds-avatar-stack'],
```
hakubo: avatars could be a property on the compent as a null
so it still blows when you don't pass it
but this way when you look at the component you see what properties it have

app/components/wds-avatar-stack.js
rybmat: @hakubo I don't think we can allow for more elements than 5 unless we'll figure out how to handle it better in scss: https://github.com/Wikia/design-system/blob/3e6f170b7379a523f3ce6c5def7db03bab063405/dist/scss/wds-components/_avatar-stack.scss#L10

latata: I would extract it to the property and leave a comment that scss file should be updated together with this value.

app/components/wds-avatar-stack.js
rybmat: done

PR #198, opened 2018-05-17
XF-555 | Introduce tests + visual tests to design-system

hakubo:


hakubo:
We decided not to purse this path for now.

PR #201, opened 2018-05-28
XW-4959 | wds-avatar great again

rybmat:
https://wikia-inc.atlassian.net/browse/XW-4959

* display default avatar image when link to user image is not provided
* tooltip over avatar
* support for user-page link after clicking on avatar
* rendering badges if permission provided

* mixin for sizing / theming avatar

@Wikia/x-wing


app/components/wds-avatar-stack.js
```diff
@@ -5,6 +5,7 @@ export default Component.extend({
 	classNames: ['wds-avatar-stack'],
 	maxStackSize: 5,
 	avatars: null,
+	translateFunc: null,
```
hakubo: as we talked
maybe we should treat i18next as peer dependency
as both mobile-wiki and discussions are using it

app/templates/components/wds-avatar/avatar-image.hbs
```diff
@@ -0,0 +1,6 @@
+{{#if src}}
+	<img src="{{src}}" alt="{{alt}}" title="{{alt}}" class="wds-avatar__image">
+{{else}}
+	<div class="wds-avatar__inner-border"></div>
+	{{svg 'wds-avatar-icon-user' class='wds-avatar__image'}}
```
hakubo: should alt and title be set here as well?

dist/scss/wds-components/_avatar-stack.scss
```diff
@@ -24,4 +12,16 @@
 		transform: translate(-40px);
 		width: auto;
 	}
+
+	.wds-avatar {
+		color: $wds-color-light-c4;
+	}
+
+	@for $i from 1 to 6 {
```
hakubo: this could be moved to .wds-avatar
`&:nth-child(#{$i})`

dist/scss/wds-components/_avatar.scss
```diff
@@ -1,5 +1,102 @@
...
+		$inner-border-width: 6px;
+	}
+
+	@include square($diameter);
+
+	a {
+		color: $color;
+	}
+
+	a:hover {
```
hakubo: please use nesting
&:hover

dist/scss/wds-components/_avatar.scss
```diff
@@ -1,5 +1,102 @@
...
+		background-color: $color-background;
+		border: solid $border-width $color;
+
+		&:hover {
+			border-color: $color-hover;
+		}
+	}
+
+	svg.wds-avatar__image {
+		fill: $color;
```
hakubo: these could be set on .wds-avatar__image
browser would just ignore it if it's not svg

dist/scss/wds-components/_avatar-stack.scss
rybmat: @hakubo I would say not to do this, in mobile-wiki we don't use avatar stack at all, so why would we load this styles?

PR #206, opened 2018-05-30
XW-4981 | Moving search logic into DS

bkoval:
* Search API is used inside DS component for search in global nav
* Requests are debounced and cached for performance
* Unified resource is used for search in Mobile-Wiki, Discussions and Oasis


package.json
```diff
@@ -79,7 +79,9 @@
     }
   },
   "dependencies": {
-    "ember-cli-babel": "6.12.0"
+    "ember-cli-babel": "6.12.0",
+    "ember-fetch": "4.0.2",
```
latata: do we need both fetch and ember-fetch?

app/components/wds-global-navigation/search.js
```diff
@@ -29,7 +59,247 @@ export default Component.extend({
...
+	 * @returns {void}
+	 */
+	setSearchSuggestionItems(suggestions = []) {
+		if (this.get('isDestroyed')) {
+			return;
+		}
+
+		const query = this.get('query'),
+			highlightRegexp = new RegExp(`(${this.escapeRegex(query)})`, 'ig'),
+			highlighted = wrapMeHelper.compute(['$1'], {
```
latata: as we use wrapMe helper only in JS it doesn't need to be helper just a simple function. unless we want to share these helper to applications and use there in hbs.

app/components/wds-global-navigation/search.js
```diff
@@ -29,7 +59,247 @@ export default Component.extend({
...
+		});
+
+		// If the query string is empty or shorter than the minimal length, return to leave the view blank
+		if (!query || query.length < this.get('queryMinimalLength')) {
+			this.set('isLoadingResultsSuggestions', false);
+		} else if (this.hasCachedResult(query)) {
+			this.setSearchSuggestionItems(this.getCachedResult(query));
+		} else {
+			this.set('isLoadingResultsSuggestions', true);
+			run.debounce(this, () => this.requestSuggestionsFromAPI(query), this.get('debounceDuration'));
```
latata: Does it work? I guess it will create new instance of debounce function each time `getSuggestions` is called. I think we should extrac `() => this.requestSuggestionsFromAPI(query)` to component method and pass this method as an argument of debounce here. But I may be wrong ;)

app/components/wds-global-navigation/search.js
```diff
@@ -1,24 +1,54 @@
...
+		this.cachedResultsQueue = [];
+
+		// key: query string, value: Array<SearchSuggestionItem>
+		this.cachedResults = {};
+	},
+
+	didInsertElement() {
+		this._super(...arguments);
+
+		run.scheduleOnce('afterRender', this, () => {
```
latata: why `after render`? Is it already rendered on didInsertElement hook?

app/components/wds-global-navigation/search.js
```diff
@@ -1,24 +1,54 @@
 import {notEmpty, empty} from '@ember/object/computed';
 import Component from '@ember/component';
-import {computed} from '@ember/object';
+import EmberObject, {computed} from '@ember/object';
 import translations from 'npm:design-system-i18n/i18n/en/design-system';
```
rybmat: if you merged latest feature branch then this is not needed, just use i18n service to translate

app/components/wds-global-navigation/search.js
```diff
@@ -29,7 +59,247 @@ export default Component.extend({
...
+			this.set('isLoadingResultsSuggestions', false);
+		} else if (this.hasCachedResult(query)) {
+			this.setSearchSuggestionItems(this.getCachedResult(query));
+		} else {
+			this.set('isLoadingResultsSuggestions', true);
+			run.debounce(this, () => this.requestSuggestionsFromAPI(query), this.get('debounceDuration'));
+		}
+	},
+
+	normalizeToUnderscore(title = '') {
```
rybmat: looks the same as in mobile-wiki, should we extract somewhere?  @hakubo ?

app/components/wds-global-navigation/search.js
rybmat: AFAIR @hakubo told me that wrapMe will be extracted to ember-fandom

app/components/wds-global-navigation/search.js
```diff
@@ -29,7 +59,247 @@ export default Component.extend({
 		return translations[this.get('model.placeholder-inactive.key')];
```
rybmat: use i18n service here and add "design-system" namespace to key

JamesHawking:
@hakubo maybe this is a good time to introduce ember-concurrency?

app/components/wds-global-navigation/search.js
bkoval: Assuming it's resolved 👍

app/components/wds-global-navigation/search.js
bkoval: Fixed!

app/components/wds-global-navigation/search.js
bkoval: I would not bother for now, we will be moving these things to a common place once we have ember-fandom in place, meanwhile this is only 1 line of meaningful code :)

app/components/wds-global-navigation/search.js
```diff
@@ -1,35 +1,309 @@
...
+			return this.get('i18n').t(`design-system:${this.get('model.placeholder-active.key')}`);
 		}
 
-		return translations[this.get('model.placeholder-inactive.key')];
+		return this.get('i18n').t(`design-system:${this.get('model.placeholder-inactive.key')}`);
 	}),
 
+	requestSuggestionsFromAPI() {
+		const query = this.get('query'),
+			uri = `${this.get('model.suggestions.url')}&query=${query}`;
```
latata: one const per var

PR #209, opened 2018-06-07
XW-4999 | search redesign & behaviour update

latata:
https://wikia-inc.atlassian.net/browse/XW-4999
@Wikia/x-wing


app/components/wds-global-navigation.js
```diff
@@ -40,7 +40,11 @@ export default Component.extend({
 		},
 
 		onSearchSuggestionChosen() {
+			// TODO
```
rybmat: what about this TODO?

style-guide/styles/wds-components/_global-navigation-search.scss
```diff
@@ -0,0 +1,166 @@
...
+			transform: rotate(180deg);
+		}
+	}
+
+	&-close {
+		display: none;
+		padding: 5px 9px;
+
+		&.wds-button.wds-is-text {
+			color: #fff;
```
rybmat: `$wds-color-white`

app/components/wds-global-navigation.js
latata: let's handle it in XW-5009

PR #226, opened 2018-06-25
IRIS-6238 | global footer as a shareable component

latata:
@Wikia/iris


app/components/wds-global-footer.js
```diff
@@ -3,4 +3,21 @@ import Component from '@ember/component';
 export default Component.extend({
 	tagName: 'footer',
 	classNames: 'wds-global-footer',
+
+	init() {
+		this._super(...arguments);
+
+		if(!this.track) {
+			throw new Error('Required `track` function for wds-global-footer component is not set');
```
rogatty: Let's use `Ember.assert()`

app/components/wds-global-footer.js
```diff
@@ -3,4 +3,21 @@ import Component from '@ember/component';
...
+		}
+	},
+
+	click(event) {
+		const elementToTrack = event.target.closest('[data-tracking-label]');
```
rogatty: Would be nice to have a boundary to prevent leaking this from the footer element.

tests/dummy/app/models/global-footer.json
```diff
@@ -2,7 +2,9 @@
 	"with-hubs": {
 		"header": {
 			"type": "link-image",
-			"image": "wds-company-logo-fandom-white",
+			"image-data": {
+        "name": "wds-company-logo-fandom-white"
```
rogatty: x

PR #228, opened 2018-06-27
XW-5051 | Partner slot for mobile

rybmat:
https://wikia-inc.atlassian.net/browse/XW-5051

@Wikia/x-wing


app/templates/components/wds-global-navigation.hbs
```diff
@@ -54,6 +54,7 @@
 	</div>
 	{{#if model.partner_slot}}
 		<div class="wds-global-navigation__partner-slot">
+			<span class="wds-global-navigation__partner-slot-horizontal-bar"></span>
```
bkoval: I'd much rather wrap partner slot into another div to make the border work than add an empty span, because the rule of thumb is to avoid empty elements in html.

If we absolutely need to use a separate element to make border top happen, let's use a div instead of span - it has display: block by default, so you won't need to add it in css.

dist/scss/wds-components/_global-navigation.scss
```diff
@@ -251,10 +251,41 @@
...
+		&-horizontal-bar {
+			display: none;
+		}
+
+		@media #{$wds-breakpoint-small-down} {
+			background: $wds-fandom-color-navy;
+			border: none;
+			height: 30px;
+			justify-content: flex-end;
+			max-width: 100vw;
```
bkoval: Is that needed, since you define width: 100vw?

PR #230, opened 2018-06-27
IRIS-5920 | Dialog component

hakubo:
You can preview it here:
http://dev-jolek/design-system/#/components/dialogs


style-guide/styles/wds-components/_dialogs.scss
```diff
@@ -0,0 +1,70 @@
...
+
+	&__actions {
+		border-top: 1px solid $wds-color-mist-gray;
+		display: flex;
+		height: 40px;
+		justify-content: flex-end;
+		padding: 5px 0 8px;
+
+		&-button {
+			font-size: $wds-typescale-size-base;
```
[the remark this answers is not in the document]
hakubo: yeah, I guess we do.
But in this case the text on buttons would too small and it would look 'ugly'

app/components/wds-dialog.js
```diff
@@ -0,0 +1,15 @@
...
+	tagName: '',
+
+	cancelText: 'Cancel',
+	okText: 'OK',
+
+	onCancel() {},
+	onOK() {},
+
+	actions: {
+		cancelBubble() {}
```
rogatty: Please add a comment explaining what is this for. It's to prevent `{{action onCancel 'curtain'}}` being triggered when `wds-dialog__wrapper` or its content are clicked, right?

rogatty:
Remember to run the build ;)

app/components/wds-dialog.js
hakubo: yes

PR #237, opened 2018-07-09
XW-5098 | add redirect param to login/logout/register urls

rybmat:
https://wikia-inc.atlassian.net/browse/XW-5098

@Wikia/x-wing


app/components/wds-global-navigation.js
```diff
@@ -1,28 +1,38 @@
...
 	],
 
 	searchIsActive: false,
 
-	searchIsAlwaysVisible: empty('model.fandom_overview'),
-
 	isSearchModalOpen: equal('currentModal', 'search'),
 	isUserModalOpen: equal('currentModal', 'user'),
+
+	signinUrl: computed('model.anon.signin', 'fastboot.isFastBoot', function() {
```
bkoval: This computed prop does not use 'fastboot.isFastBoot' -> is adding it to dependencies here intentional?

app/components/wds-global-navigation.js
```diff
@@ -31,6 +41,14 @@ export default Component.extend({
 		assert('Required property `model` is not set', this.model);
 	},
 
+	getRedirectUrl() {
```
bkoval: Maybe this could be a computed property based on 'fastboot.isFastBoot'? Then it would recalculate automatically both on fastboot and front-end.

PR #241, opened 2018-07-12
XW-5120 | Community bar uses wgSitename

bkoval:
@Wikia/x-wing @mklucsarits


dist/scss/wds-components/global-footer/_common.scss
```diff
@@ -21,6 +21,11 @@ $footer-line-height: 1.3em;
 		padding: 0;
 	}
 
+	&__header-wrapper {
```
bkoval: This is something extra (added in one of the previous commits but not built).

PR #249, opened 2018-07-16
IRIS-6266 | When a notification is clicked then mark it as read

rogatty:
https://wikia-inc.atlassian.net/browse/IRIS-6266
https://github.com/Wikia/app/pull/15718

cc @Wikia/iris


rogatty:
I've tested it in discussions but will wait with a PR until this is released.

app/components/code-sample.js
```diff
@@ -1,14 +0,0 @@
-import Component from '@ember/component';
-import { registerHljsLanguage } from '../utils/beautify';
-
-export default Component.extend({
```
rogatty: It's not used, there is https://github.com/Wikia/design-system/blob/master/tests/dummy/app/components/code-sample.js

PR #250, opened 2018-07-17
IRIS-6402 | Avatar stack fixes

rogatty:
https://wikia-inc.atlassian.net/browse/IRIS-6402

- We have cases in Discussions where there are hundreds of users that should be accounted for in the overflow (+99) text but we only have 5 avatars. Let's add an argument that allows to override this number.
- Use `margin-left` instead of `transform: translate` to make `wds-avatar-stack` container size correct and avoid this:
screenshot from 2018-07-17 13-15-44
- Allow to hide the avatar stack overflow text
- Remove default JPG avatar from on-site-notifications and instead use a DS icon which is already handled by `wds-avatar`
- Bonus: fix `component-demo` which was rendering its own source code


app/components/wds-avatar-stack.js
```diff
@@ -5,16 +5,22 @@ export default Component.extend({
...
 	avatars: null,
+	overrideCount: null,
 
-	overflow: computed('avatars', function() {
-		const count = this.get('avatars').length || 0;
+	count: computed('avatars', 'overrideCount', function () {
+		return this.overrideCount || this.get('avatars.length') || 0;
+	}),
+
+	overflow: computed('avatars', function () {
```
rybmat: shouldn't this computed property be dependent on `count` as well?

PR #253, opened 2018-07-19
IRIS-6405 | Use sendBeacon to mark notification as read on page unload

rogatty:
https://wikia-inc.atlassian.net/browse/IRIS-6405
https://github.com/Wikia/app/pull/15736

@Wikia/iris

### DIFF
--- a/app/services/wds-on-site-notifications.js
+++ b/app/services/wds-on-site-notifications.js
@@ -80,12 +80,16 @@ export default Service.extend({
 		this.get('model').markAllAsRead();
 	},
 
-	markAsRead(notification) {
-		this.get('model').markAsRead(notification);
+	markAsRead(notification, willUnloadPage = false) {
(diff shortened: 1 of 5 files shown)


app/services/wds-on-site-notifications.js
```diff
@@ -80,12 +80,16 @@ export default Service.extend({
...
+		return this.get('model').markAsRead(notification, willUnloadPage);
 	},
 
 	getUnreadCount() {
 		return this.get('model.unreadCount');
-	}
+	},
 
+	goToDestination(notification) {
+		// TODO check if it's a local url and use transitionTo or delegate to app
```
rybmat: what about this TODO?

rogatty: It's not a part of this ticket as it's already like that on production.
See https://wikia-inc.atlassian.net/browse/MAIN-16952

PR #254, opened 2018-07-25
XW-5119 | New Typography

bkoval:
@Wikia/x-wing


dist/scss/wds-typography/placeholders/_line-height-scale.scss
```diff
@@ -0,0 +1,47 @@
...
+
+%wds-line-height-4xl {
+	line-height: 52px;
+}
+
+%wds-line-height-paragraph {
+	line-height: 28px;
+}
+
+%wds-line-height-short-paragraph {
```
hakubo: it's the same as wds-line-height-paragraph
why do we need both?

tests/dummy/app/components/font-weight-line.js
```diff
@@ -0,0 +1,5 @@
+import Component from '@ember/component';
+
+export default Component.extend({
+	classNames: 'typescale-line'
```
hakubo: tagName: ''
and add the wrapping div to template
cleaner approach
one that will be soon default

dist/scss/wds-typography/placeholders/_text-style.scss
```diff
@@ -0,0 +1,26 @@
...
+
+%wds-text-transform-lowercase {
+	text-transform: lowercase;
+}
+
+%wds-text-transform-none {
+	text-transform: none;
+}
+
+%wds-midlight {
```
vforge: It won't work as expected for multi-line, you need to add `box-decoration-break: clone;`
see https://github.com/Wikia/f2/search?q=box-decoration-break&unscoped_q=box-decoration-break
see https://developer.mozilla.org/en-US/docs/Web/CSS/box-decoration-break

rybmat: @vforge we tested it and it works: 
screen shot 2018-07-25 at 15 47 23
the only case we are aware it does not work and `box-decoration-break: clone;` doesn't help either is when parent element uses `display: flex`

rybmat: btw, we should use % in the gradient instead of px to support all font sizes

vforge: @rybmat There's 3px padding on the left on the 1st line and not on the right side of the 1st line. There's no padding on the 2nd line and on the left side on the 3rd line. The right side on the 3rd line has a padding.

rybmat: obviously I'm blind ;)

bkoval: I added the rule, thanks @vforge

PR #262, opened 2018-08-29
XW-5110 | fix Global -> glocal nav transition

rybmat:
https://wikia-inc.atlassian.net/browse/XW-5110

@Wikia/x-wing


app/components/wds-global-navigation.js
```diff
@@ -71,13 +71,15 @@ export default Component.extend({
 	},
 
 	scrollHandler() {
-		const globalNavigationHeight = 55;
+		const communityBar = document.querySelector('.wds-community-bar');
```
bkoval: Could we evaluate the selector once (e.g. in didInsertElement)? If we leave it here, it will mean a DOM operation each time scroll handler is fired.

PR #265, opened 2018-09-05
IRIS-6490 | wds-input styles moved from react-design-system

latata:
https://wikia-inc.atlassian.net/browse/IRIS-6490


xkxd:
why did it update `docs/assets/vendor.js` if only CSS changed?

PR #278, opened 2018-10-15
 Add support for icons on the right side of a button

rogatty:
Equal rights for right-side icons! The dropdown chevron will be moved by 1px as this would be overriden anyway:
```
.wds-dropdown__toggle-chevron:not(:only-child) {
    margin-left: 5px;
}
```
No one will notice :see_no_evil: 

@Wikia/iwing


dist/scss/wds-components/buttons/_common.scss
```diff
@@ -56,13 +56,17 @@
 		border: 0;
 	}
 
-	.wds-icon:first-child {
+	.wds-icon {
 		align-self: center;
 		pointer-events: none;
 
-		&:not(:only-child) {
+		&:first-child:not(:only-child) {
```
hakubo: how about not repeating a bit of css selector?
```
:not(only-child) {
  &:first-child {}
  &:last-child {}
}
```

PR #282, opened 2018-10-25
IW-1106 | fix badges

xkxd:
https://wikia-inc.atlassian.net/browse/IW-1106

@Wikia/iwing


app/components/wds-avatar/avatar-badge.js
```diff
@@ -2,6 +2,11 @@ import Component from '@ember/component';
 import { computed } from '@ember/object';
 import { inject as service } from '@ember/service';
 
+const userRoleMap = {
```
hakubo: missing tests for this

hakubo: https://github.com/Wikia/design-system/blob/master/tests/integration/components/wds-avatar-test.js#L28

PR #283, opened 2018-10-25
support SPA links in Githubpages

hakubo:
https://github.com/rafrex/spa-github-pages

This will add
- support for normal paths
- fix loading iframes on firefox


hakubo:
yes!

PR #309, opened 2018-12-13
IW-1421 | fix submitting search query

mszabo-wikia:
When user submits search form using Enter key, they should be redirected to the search results page.

https://wikia-inc.atlassian.net/browse/IW-1421


tests/integration/components/global-navigation/search-test.js
```diff
@@ -79,27 +79,26 @@ module('Integration | Component | global-navigation/search', function (hooks) {
...
-	// 		onSearchToggleClicked=onSearchToggleClicked
-	// 	}}`);
-
-	// 	const searchInputSelector = '.wds-global-navigation__search-input';
-
-	// 	await click('.wds-global-navigation__search-toggle-icon');
-	// 	await fillIn(searchInputSelector, 'query');
-	// 	await triggerKeyEvent('.wds-global-navigation__search-container', 'keydown', 13);
-	// });
+	test('search input results are shown after click on Enter', async function (assert) {
```
hakubo: why it started to work?

app/components/global-navigation/search.js
```diff
@@ -355,6 +355,9 @@ export default Component.extend({
 				this.inputField.blur();
 
 				this.closeSearch();
+			} else if (this.goToSearchResults) {
```
hakubo: it actually works for me on prod and preview. Is the bug on sandbox somewhere?

mszabo-wikia: The bug is in current dev branch, reproducible if you run it e.g. on devbox

tests/integration/components/global-navigation/search-test.js
mszabo-wikia: it works with fix in this PR. uncommenting it without fix leads to it not working.

PR #312, opened 2018-12-20
Fix search suggestions

hakubo:
- also add classes helper (with tests!)
- fix search suggestions on demo page


app/helpers/classes.js
```diff
@@ -0,0 +1,7 @@
+import { helper } from '@ember/component/helper';
+
+export function classes(classList) {
```
rybmat: https://github.com/JedWatson/classnames
can we use this? This way it will be the same as in react stack

hakubo: ha good one. For now I don't think we need such tool here
but I've renamed classes to classnames - I like that name better. and API of ot is very similar now

PR #317, opened 2019-01-07
IW-1341 | Close navigation when sub-level title is clicked

mszabo-wikia:
If an user clicks a level 2 item title in mobile-wiki community header, the navigation bubble remains visible even though the page has changed. The navigation should always be closed if it triggers a page change.

https://wikia-inc.atlassian.net/browse/IW-1341

@Wikia/iwing

### DIFF
--- a/app/components/community-bar/local-navigation-dropdown.js
+++ b/app/components/community-bar/local-navigation-dropdown.js
@@ -73,12 +73,7 @@ export default Component.extend({
 		}
 	},
 
-	onThirdLevelSelected(event) {
-		this.toggleNavigation(false);
-		this.onLinkClicked(event);
-	},
-
-	onSubLevelTitleClicked(event) {
+	onNavigationItemLinkClicked(event) {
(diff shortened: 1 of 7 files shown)


mszabo-wikia:
I only just noticed https://github.com/Wikia/design-system/pull/316, probably worth combining the two

app/components/community-bar/local-navigation-dropdown.js
```diff
@@ -78,6 +78,11 @@ export default Component.extend({
 		this.onLinkClicked(event);
 	},
 
+	onSubLevelTitleClicked(event) {
```
hakubo: hmm this looks exactly the same as onThirdLevelSelected do we need another method?

hakubo: maybe rename the method so it could apply to more places?

hakubo:
@mszabo-wikia feel free to merge this https://github.com/Wikia/design-system/pull/316 to this PR.

PR #321, opened 2019-01-15
IW-1501 | default avatar's inner border refactored

ustczanka:
https://wikia-inc.atlassian.net/browse/IW-1501


ustczanka:
I'm closing this PR, we don't want to merge it because default avatar's issue was solved in IW-1261.

PR #326, opened 2019-02-04
IW-1022 | Full-width Global Navigation

latata:


app/templates/components/global-navigation.hbs
```diff
@@ -1,137 +1,135 @@
-<div class="wds-global-navigation__content-container" ...attributes>
-  <div class="wds-global-navigation__content-bar-left">
+<div class="wds-global-navigation__content-bar-left">
```
latata: I only removed `wds-global-navigation__content-container` wrapper in this files

tests/dummy/app/services/i18n.js
```diff
@@ -8,7 +8,7 @@ export default Service.extend({
 	rootURL: ENV.rootURL,
 
 	initialize(language) {
-		return fetch(`${this.rootURL}/assets/i18n/${language}/design-system.json`)
+		return fetch(`${this.rootURL}assets/i18n/${language}/design-system.json`)
```
latata: Reference page didn't want to run because there were to slashes at the beginning of this path.

app/templates/components/dropdown-content.hbs
```diff
@@ -1,5 +1,6 @@
 <div
   class={{classnames
+    @class
```
latata: It fixes the bug on production - ember global nav search suggestions dropdown content doesn't have right styles applied.

PR #327, opened 2019-02-08
IW-1498 | Exposing data for search suggestions tracking

bkoval:
@Wikia/iwing


app/components/global-navigation/search.js
```diff
@@ -34,6 +35,7 @@ export default Component.extend({
 	},
 	/* eslint-disable ember/avoid-leaking-state-in-ember-objects */
 	isEmptyQuery: empty('state.query'),
+	searchId: null,
```
rybmat: in suggestions events there is no such thing like `searchId`, you should name it `suggestionId `

app/components/global-navigation.js
```diff
@@ -173,6 +173,11 @@ export default Component.extend({
 			this.onSearchSuggestionChosen(suggestion);
 		},
 
+		onSearchSuggestionsImpression(suggestions, searchId) {
+			// eslint-disable-next-line no-console
+			console.info('Suggestions impression', suggestions, 'with searchId', searchId);
```
rybmat: Why we need console.log here?

bkoval: This is definitely not the place for console.log, it was supposed to land in a controller of the dummy app :)

app/components/global-navigation/search.js
```diff
@@ -92,7 +94,9 @@ export default Component.extend({
 					return response.json().then((data) => {
 						const suggestions = data.suggestions.map((suggestion) => {
 							return EmberObject.create({
-								title: suggestion
+								title: suggestion,
+								articleId: data.ids && data.ids[suggestion],
+								query: data.query,
```
rybmat: I think this field should be named `value`

PR #342, opened 2019-04-03
IW-1808 | White global nav for wikia.org

bkoval:
@Wikia/iwing


app/templates/components/global-footer-fandom.hbs
```diff
@@ -0,0 +1,57 @@
+{{#if @model.header}}
```
bkoval: The whole content of this file was moved from global-footer.hbs

dist/scss/wds-components/_global-navigation-search.scss
```diff
@@ -234,6 +234,59 @@ $toggle-active-margin-right: 17px;
 	}
 }
 
+.wds-is-wikia-org {
```
latata: Can we move all of wikia.org related styles to separate files and load them conditionally only on wikia.org communities?

app/components/global-footer-fandom.js
```diff
@@ -0,0 +1,8 @@
+import Component from '@ember/component';
+import { oneWay } from '@ember/object/computed';
+import track from '../utils/wds-track';
+
+export default Component.extend({
```
xkxd: tbh we could have this as a `template-only` component

app/components/global-footer-wikia-org.js
```diff
@@ -0,0 +1,7 @@
+import Component from '@ember/component';
+
+export default Component.extend({
+	tagName: 'footer',
+	classNames: 'wds-global-footer-wikia-org',
+	year: new Date().getYear() + 1900,
```
xkxd: getFullYear

app/components/global-footer-wikia-org.js
```diff
@@ -0,0 +1,7 @@
+import Component from '@ember/component';
+
+export default Component.extend({
+	tagName: 'footer',
+	classNames: 'wds-global-footer-wikia-org',
+	year: new Date().getFullYear(),
```
latata: It may cause an issue at the turn of the years. I think `new Date().getFullYear()` will be run only once per deployment on fastboot. But I think no one will notice such issue. ;) Just wanted to point it out because in some situations it may be more prominent.

dist/scss/wds-components/_global-navigation-wikia-org.scss
```diff
@@ -0,0 +1,164 @@
+@import 'global-navigation-search';
```
latata: Please make sure that won't duplicate styles in the styles bundle.

dist/scss/wds-components/_global-navigation-search.scss
bkoval: https://wikia-inc.atlassian.net/browse/IW-1904

PR #346, opened 2019-04-12
IW-1869 | main page label

latata:
TODO:
* upload string to crowdin and create a ticket for LOC team
* download translated strings
* install new version of DS in all clients which have local navigation


bkoval:
The check is failing due to long path problem on Jenkins, unrelated to the code.

PR #352, opened 2019-06-10
IW-2107 | license string

rybmat:
https://wikia-inc.atlassian.net/browse/IW-2107

Adding license description copy

@Wikia/iwing


docs/assets/i18n/de/design-system.json
```diff
@@ -42,7 +43,7 @@
   "global-footer-create-wiki-link-start-wikia": "Wiki erstellen", 
   "global-footer-fandom-app-description": "So hast du deine Lieblings-Communitys immer dabei und kannst nichts verpassen", 
   "global-footer-fandom-app-header": "Die FANDOM-App", 
-  "global-footer-fandom-overview-header": "Explore",
```
rybmat: I have no idea what happened here, when I'm on master locally, all of these strings are translated already...

PR #373, opened 2019-12-31
Fix wds-icons-external-small

vqbc:
Remove a mask that causes issues with global footer icons when WDSIcons dev script is imported


mszabo-wikia:
Hi, can you elaborate a bit on what problem this mask is causing and how it can be reproduced?

Thanks for your contribution :)

vqbc:
The error previously caused the app icons in the global footer to display incorrectly (sorry, I don't think I have a screenshot) when the Dev Wiki WDSIcons script was imported, usually for use in another script. I believe that the issue has been resolved with a temporary measure in the script in question, but fixing the svg icon at the source is still a good idea.

PR #378, opened 2020-01-27
IW-2879 | increase selector specificity for theming links in dropdowns

rybmat:
https://wikia-inc.atlassian.net/browse/IW-2879

@Wikia/iwing


nikodamn:
Tests passed, Jenkins job failed due to some other problems.

PR #381, opened 2020-02-05
Mixins fixes

ustczanka:


dist/scss/wds-mixins/_dialog-theming.scss
```diff
@@ -8,16 +8,16 @@
 	}
 
 
-	&__curtain {
+	.wds-dialog__curtain {
```
xkxd: why is that?

ustczanka: because it only make sense for classes like `.wds-dialog__curtain` to be affected by this mixin. If somebody puts a call of this mixin inside some container with other class name, it wouldn't work (in its previous form).

PR #385, opened 2020-02-12
PLATFORM-4570 | CNW placeholder is out of its place in some languages

maaartyyynaa:
https://wikia-inc.atlassian.net/browse/PLATFORM-4570


dist/scss/wds-components/inputs/_input.scss
```diff
@@ -63,6 +63,14 @@
 		color: $wds-color-blue-gray;
 	}
 
+	.is-empty > &__field ~ &__label {
+		overflow: hidden;
+		height: 1rem;
+		text-overflow: ellipsis;
```
harnash: I'm not totally sure this is a good approach. We will truncate the text in the hint which kind of defeats its purpose.

latata: this selector is already defined in line 56. let's move these styles there

PR #412, opened 2020-09-30
IW-4318 | Ember AC notifications

kasperekt:
https://wikia-inc.atlassian.net/browse/IW-4318


latata (approved):

app/components/on-site-notifications/notification-card.js
```diff
@@ -175,6 +205,18 @@ export default Component.extend(
...
 
+		isArticleCommentReply(type) {
+			return type === notificationTypes.articleCommentReply;
+		},
+
+		isArticleCommentAtMention(type) {
+			return type === notificationTypes.articleCommentAtMention;
+		},
+
+		isArticleCommentReplyAtMention(type) {
```
iraminius, line 216: Do you think we can get rid of those `is` functions? It makes this file bigger and I don't think it is more readable this way. I know you just followed the pattern that was here but just wondering if we can get rid of this and just use switch statements for `iconName` and `textWithHtml`

kasperekt: I suppose we could. That'll be nicer.

PR #422, opened 2020-10-27
SER-4349 A/B/C adding drop down on Global Nav with a search scope [mweb]

mmeller-wikia:
https://wikia-inc.atlassian.net/browse/SER-4349


app/components/global-navigation/search.js
```diff
@@ -69,36 +69,49 @@ export default Component.extend({
 
 	requestSuggestionsFromAPI() {
 		const { query } = this.state;
-		const uri = `${this.get('model.suggestions.url')}&query=${query}`;
-
+		const scope = this.get('searchScope');
```
latata, line 72: Let's add default value for scope e.g. `const scope = this.get('searchScope') || '';`
I think that without it we could end up with requests with param `&scope=undefined`

mmeller-wikia: I set default fallback to this-wiki search

dist/scss/wds-components/_global-navigation-search.scss
```diff
@@ -7,6 +7,20 @@ $toggle-active-margin-right: 17px;
 	height: 100%;
 	justify-content: flex-end;
 
+	&-scope-dropdown {
+		&::before, &::after {
+			z-index: 150 + 1;
+		}
+
+		&::after {
+			bottom: -13px !important;
```
mmeller-wikia, line 16: I don't like this `!important` but i lost my patience for this style

dist/scss/wds-components/_global-navigation-search.scss
```diff
@@ -7,6 +7,20 @@ $toggle-active-margin-right: 17px;
 	height: 100%;
 	justify-content: flex-end;
 
+	&-scope-dropdown {
+		&::before, &::after {
+			z-index: 150 + 1;
```
lomiarz, line 12: how about to just write 151 🔢

app/components/global-navigation/search.js
```diff
@@ -56,6 +56,17 @@ export default Component.extend({
 
 		// key: query string, value: Array<SearchSuggestionItem>
 		this.cachedResults = {};
+		this.setScopeMessage(this.getScope());
+	},
+
+	setScopeMessage(scope) {
+		if (scope === 'cross-wiki') {
+			this.set('model.placeholder-active.key', 'global-navigation-search-placeholder-all-wikis');
```
lomiarz, line 64: small thing, please try to keep consistent names, here you have `cross-wiki` vs `all-wikis`

lomiarz: also you could introduce some enumeration for those hardcoded keys

app/components/global-navigation/search.js
```diff
@@ -377,9 +423,24 @@ export default Component.extend({
 		}
 	},
 
+	getScope() {
+		return this.get('searchScope') === 'cross-wiki' ? 'cross-wiki' : 'internal';
```
lomiarz, line 427: what is a benefit from that function ? searchScope can have only two values

app/templates/components/global-navigation/search.hbs
```diff
@@ -84,6 +117,9 @@
                 data-tracking-label={{model.suggestions.tracking-label}}
               >
                 {{{suggestion.text}}}
+                {{#if suggestion.sitename}}
+                  <span style="float: right">{{{suggestion.sitename}}}</span>
```
lomiarz, line 121: IMHO you could add a class name and describe style in css file

kejwmen (commented):
Please take a look at @lomiarz comments

kejwmen (approved):

PR #425, opened 2020-11-03
CAKE-6064 | Add Message Wall Content Types to On-Site Notifications

cschucode:
https://wikia-inc.atlassian.net/browse/CAKE-6064

Add Message Wall content types to be consumed by mobile-wiki.

* Several other unrelated files were changed when building the files for production.

Who may be interested?
@Wikia/cake 
@Wikia/iwing


JoshuaRogan (approved):
LGTM given a smoke test with the other file changes.

app/components/on-site-notifications/notification-card.js
```diff
@@ -295,6 +303,101 @@ export default Component.extend(
...
+		getMessageWallPostBody(model) {
+			let wallOwner = model.get('metadata') && model.get('metadata.username');
+		
+			if (!wallOwner) {
+				wallOwner = this.getMessageWallOwner(model.get('uri'));
+			}
+		
+			const isOwnWall = wallOwner === this.usernameMarkup;
+			const args = {
+				postTitle: this.postTitleMarkup,
```
mklucsarits, line 315: This should be set to `model.get('title')`. There is a `this.titleMarkup` computed value which is for bolding the title, but that's not what we want. For the wall notifications, the title is bolded in the localized strings where necessary. And `this.postTitleMarkup` is undefined.

app/components/on-site-notifications/notification-card.js
```diff
@@ -295,6 +303,101 @@ export default Component.extend(
...
+		getMessageWallReplyBody(model) {
+		let wallOwner = model.get('metadata') && model.get('metadata.username');
+	
+		if (!wallOwner) {
+			wallOwner = this.getMessageWallOwner(model.get('uri'));
+		}
+	
+		const isOwnWall = wallOwner === this.usernameMarkup;
+		const args = {
+			postTitle: this.postTitleMarkup,
```
mklucsarits, line 339: Same comment as above.

app/components/on-site-notifications/notification-card.js
```diff
@@ -295,6 +303,101 @@ export default Component.extend(
...
+				return this.getTranslatedMessage('notifications-own-wall-post', args);
+			}
+		
+			args.firstUser = this.getPossiblyAnonActorName(model);
+			// "{firstUser} left a <b>new message</b> on {wallOwner}'s wall <br><br> {postTitle}"
+			return this.getTranslatedMessage('notifications-wall-post', args);
+		},
+
+		getMessageWallReplyBody(model) {
+		let wallOwner = model.get('metadata') && model.get('metadata.username');
```
mklucsarits, line 331: Indentation is off for this whole function block.

app/components/on-site-notifications/notification-card.js
```diff
@@ -295,6 +303,101 @@ export default Component.extend(
...
+
+		getMessageWallPostRemovedBody() {
+			// "<b>Message removed</b> from your wall <br><br> {postTitle}"
+			return this.getTranslatedMessage('notifications-own-wall-post-removed', {
+				postTitle: this.postTitleMarkup,
+			});
+		},
+
+		getMessageWallOwner(url) {
+		const regex = /\/Message_Wall:(.+?)([?#/].*)?$/i;
```
mklucsarits, line 385: Indentation needs fixing here and in the next function as well.

app/components/on-site-notifications/notification-card.js
```diff
@@ -295,6 +303,101 @@ export default Component.extend(
...
+			args.number = model.get('totalUniqueActors') - 1;
+	
+			if (isOwnWall) {
+			args.user = this.getPossiblyAnonActorName(model);
+			// "{user} and {number} other users <b>replied</b> to a message on your wall <br><br> {postTitle}",
+			return this.getTranslatedMessage('notifications-own-wall-reply-multiple-users', args);
+			}
+	
+			args.firstUser = this.getPossiblyAnonActorName(model);
+			args.secondUser = model.get('contentCreatorName');
```
mklucsarits, line 353: `model.get('contentCreatorName')` could be empty if it's an anonymous user. We need to check for that and use the anon user string instead. Looks like this is broken/incorrect in unified-platform but should be fixed here too.

app/components/on-site-notifications/notification-card.js
```diff
@@ -295,6 +303,101 @@ export default Component.extend(
...
+	
+		if (model.get('contentCreatorName') === this.usernameMarkup) {
+			// Current user's own message
+			args.user = this.getPossiblyAnonActorName(model);
+			// "{user} <b>replied</b> to your message <br><br> {postTitle}"
+			return this.getTranslatedMessage('notifications-wall-reply-own-message', args);
+		}
+	
+		args.firstUser = this.getPossiblyAnonActorName(model);
+		args.secondUser = model.get('contentCreatorName');
```
mklucsarits, line 372: Also could be anon/empty string.

app/components/on-site-notifications/notification-card.js
cschucode: Updated.

app/components/on-site-notifications/notification-card.js
cschucode: Updated.

app/components/on-site-notifications/notification-card.js
cschucode: Updated.

app/components/on-site-notifications/notification-card.js
cschucode: Using `model.get('title')`.

app/components/on-site-notifications/notification-card.js
cschucode: Yep. Makes sense. Updated.

app/components/on-site-notifications/notification-card.js
cschucode: Same as above. Updated.

app/components/on-site-notifications/notification-card.js
```diff
@@ -295,6 +303,94 @@ export default Component.extend(
...
+			}
+
+			const isOwnWall = wallOwner === this.usernameMarkup;
+			const args = {
+				postTitle: model.get('title'),
+				wallOwner,
+			};
+
+			if (isOwnWall) {
+				// "{user} left a <b>new message</b> on your wall <br><br> {postTitle}",
```
mklucsarits, line 320: Not sure if you intended to remove all these but if so you missed one.

mklucsarits (approved):
This is looking good to me. We just need to do some manual testing to make sure of things.

app/models/wds-on-site-notifications/wds-on-site-notification.js
```diff
@@ -77,9 +78,29 @@ export default EmberObject.extend({
...
+			case 'message-wall-reply-notification':
+				return notificationTypes.messageWallReply;
+			case 'wall-post-removed-notification':
+				return notificationTypes.messageWallPostRemoved;
 		}
 	},
 
+	getMetadata(notification) {
+		let metadata =
+			typeof notification.metadata === 'undefined' || notification.metadata === null ? null : notification.metadata;
```
kasperekt, line 92: Why not `let metadata = notification.metadata || null` or better (if we have proper settings) `let metadata = notification.metadata ?? null`

app/models/wds-on-site-notifications/wds-on-site-notification.js
```diff
@@ -77,9 +78,29 @@ export default EmberObject.extend({
...
+		if (typeof metadata === 'string') {
+			try {
+				metadata = JSON.parse(metadata);
+			} catch (e) {
+				// continue regardless of error
+			}
+		}
+		// Caveat: if the metadata was not serialized JSON, a string could be returned instead of an object
+		return metadata;
```
kasperekt, line 101: Is it handled later as string and object?

app/components/on-site-notifications/notification-card.js
```diff
@@ -295,6 +303,93 @@ export default Component.extend(
...
 		},
 
+		getMessageWallPostBody(model) {
+			let wallOwner = model.get('metadata') && model.get('metadata.username');
+
+			if (!wallOwner) {
+				wallOwner = this.getMessageWallOwner(model.get('uri'));
+			}
+
+			const isOwnWall = wallOwner === this.usernameMarkup;
```
kasperekt, line 313: Is username unique? Maybe we should use IDs here.

app/components/on-site-notifications/notification-card.js
```diff
@@ -295,6 +303,93 @@ export default Component.extend(
 			});
 		},
 
+		getMessageWallPostBody(model) {
+			let wallOwner = model.get('metadata') && model.get('metadata.username');
```
kasperekt, line 307: Metadata can be a string.

app/models/wds-on-site-notifications/wds-on-site-notification.js
```diff
@@ -77,9 +78,29 @@ export default EmberObject.extend({
...
 	},
 
+	getMetadata(notification) {
+		let metadata =
+			typeof notification.metadata === 'undefined' || notification.metadata === null ? null : notification.metadata;
+		if (typeof metadata === 'string') {
+			try {
+				metadata = JSON.parse(metadata);
+			} catch (e) {
+				// continue regardless of error
```
kasperekt, line 97: Maybe it's better to return `undefined`, empty object or object with default values, here in catch block

kasperekt (changes requested):

app/components/on-site-notifications/notification-card.js
mklucsarits: It won't be for this notification.

app/components/on-site-notifications/notification-card.js
mklucsarits: Username is unique across MW.

app/models/wds-on-site-notifications/wds-on-site-notification.js
mklucsarits: It depends on the notification type. The metadata is 100% developer controlled as to whether it's a string or object for the notification they're creating.

app/models/wds-on-site-notifications/wds-on-site-notification.js
mklucsarits: We'd like it to match the behavior in unified-platform https://github.com/Wikia/unified-platform/blob/master/extensions/fandom/DesignSystem/scripts/DesignSystemGlobalNavigationOnSiteNotificationsController.js#L91-L101

mklucsarits: `null` is intentional to mean: this field was handled and there's nothing in it. If it's `undefined`, this could mean either it was set to `undefined` or the field was never handled by the code, and you won't know which one.

app/components/on-site-notifications/notification-card.js
cschucode: Looks like we can use `const currentUsername = this.wdsOnSiteNotifications.currentUser.name;`

app/components/on-site-notifications/notification-card.js
cschucode: This has also been updated to reflect pandora change `model.get('metadata.wallOwnerName')`.

app/models/wds-on-site-notifications/wds-on-site-notification.js
cschucode: I agree this is consistent with UCP. The comment was added to keep the linter from complaining.

app/components/on-site-notifications/notification-card.js
```diff
@@ -295,6 +303,98 @@ export default Component.extend(
 			});
 		},
 
+		getMessageWallPostBody(model) {
+			let wallOwner = model.get('metadata') && model.get('metadata.wallOwnerName');
```
mklucsarits, line 307: `model.get` is null-safe, so you can use `model.get('metadata.wallOwnerName')` by itself.

i18n/fi/design-system.json
```diff
@@ -162,6 +162,7 @@
   "wds-avatar-badges-staff-tooltip": "FANDOM-henkilökunta", 
   "wds-avatar-badges-vstf-tooltip": "VSTF", 
   "wds-fandom-homepage-aria-label": "Fandom homepage", 
-  "wds-gamepedia-homepage-aria-label": "Gamepedia homepage", 
-  "wds-wikiaorg-homepage-aria-label": "Wikia.org homepage"
-}
\ No newline at end of file
+  "wds-gamepedia-homepage-aria-label": "Gamepedia homepage",
+  "wds-wikiaorg-homepage-aria-label": "Wikia.org homepage",
+  "username-anonymous": "A FANDOM User"
```
mklucsarits, line 167: These should use "Fandom" instead but we have to be careful if manually changing them, as I don't know how that works for non-Latin alphabets like Russian and the Asian languages.

cschucode: I'm going to remove this for now and handle translations in a separate ticket.

app/models/wds-on-site-notifications/wds-on-site-notification.js
cschucode: Yep. that works too. making the change.

kasperekt (approved):

PR #426, opened 2020-11-04
Revert "Revert "CAKE-5920 Update some notifications to include wall owner's name""

mklucsarits:
Reverts Wikia/design-system#424


mklucsarits:
Already approved in the original PR and unchanged since then.

PR #431, opened 2020-11-12
SER-4349 A/B/C adding drop down on Global Nav with a search scope [mweb]

mmeller-wikia:
https://wikia-inc.atlassian.net/browse/SER-4349

### DIFF
--- a/app/templates/components/global-navigation/search.hbs
+++ b/app/templates/components/global-navigation/search.hbs
@@ -109,7 +109,7 @@
             <li
               class="{{if (equal index selectedSuggestionIndex) "wds-is-selected"}} wds-global-navigation__search__suggestion"
               onmouseenter={{action (mut selectedSuggestionIndex) index}}
-              {{action "onSearchSuggestionClick" index}}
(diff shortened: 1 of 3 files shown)


app/templates/components/global-navigation/search.hbs
```diff
@@ -109,7 +109,7 @@
             <li
               class="{{if (equal index selectedSuggestionIndex) "wds-is-selected"}} wds-global-navigation__search__suggestion"
               onmouseenter={{action (mut selectedSuggestionIndex) index}}
-              {{action "onSearchSuggestionClick" index}}
+              {{action "onSearchSuggestionClick" index preventDefault=(equal (equal searchScope "cross-wiki") false)}}
```
rybmat, line 112: I'd still check that condition in the action handler to not send request to current wiki if not needed

rybmat: Right now AFAIK, both things will happen: browser will redirect you to the link you click, but in the background unnecessary request (that will end up with 404 and response will not be handled) will be sent

rybmat (approved):

PR #432, opened 2020-11-13
cake-6121 | Add talk page notification

cschucode:
https://wikia-inc.atlassian.net/browse/CAKE-6121

Adds Talk Page Message Notification to DS.


docs/assets/i18n/ko/design-system.json
```diff
@@ -118,6 +118,7 @@
   "hours-ago": "{hours}h", 
   "license-description": "Community content is available under {license} unless otherwise noted.", 
   "minutes-ago": "{minutes}m", 
+  "notifications-anon-user": "A Fandom user",
```
mklucsarits, line 121: It would useful to have this in the design system but not coupled to notifications, so I would vote for a more generalized key name.

cschucode: This was added in https://github.com/Wikia/design-system/pull/430 and wasn't readily available in my previous PR. I noticed the duplication with `username-anonymous` and opted for this since we're also using it for notifications.

mklucsarits (approved):

PR #438, opened 2020-12-10
SER-4524 [Mobile] Search scope toggle

mmeller-wikia:


slayful (approved):

latata (approved):

app/components/global-navigation/search.js
```diff
@@ -411,6 +414,10 @@ export default Component.extend({
 		// ENTER key
 		} else if (keyCode === 13) {
 			if (this.selectedSuggestionIndex !== -1) {
+				if (this.getScope() === 'cross-wiki') {
+					window.location.replace(this.suggestions[this.selectedSuggestionIndex].uri);
```
kasperekt, line 418: Why `replace` instead on `push`? Is it intentional?

app/components/global-navigation/search.js
```diff
@@ -430,6 +437,15 @@ export default Component.extend({
 		return this.get('searchScope') === 'cross-wiki' ? 'cross-wiki' : 'internal';
 	},
 
+	getDefaultScope() {
+		const savedSearchScope = Cookies.get('DEFAULT_SEARCH_SCOPE');
```
lomiarz, line 441: how about just `return Cookies.get('DEFAULT_SEARCH_SCOPE') || 'internal'`

app/components/global-navigation/search.js
```diff
@@ -430,6 +437,15 @@ export default Component.extend({
...
+		if (savedSearchScope === 'cross-wiki' || savedSearchScope === 'internal') {
```
lomiarz, line 442: did you consider to use some enum/constants for cross-wiki and internal scope ?

app/components/global-navigation/search.js
mmeller-wikia: We need full reload (redirect to different wiki) and this is just the first solution that worked for me
I can't find `window.location.push` function, so if you can provide snippet i would be grateful :)

app/components/global-navigation/search.js
mmeller-wikia: I would stick with this little validation. Just because different value will break nav-search

lomiarz (approved):

